PR #56, opened 2014-01-24
Add documentation for Red Hat install.

mmcclimon:
Hi Jon-

I wrote up a guide to installing on Red Hat, after many trials and tribulations. I think it's pretty self-explanatory, but you'll want to make sure everything is acceptable. In particular, Red Hat required a slightly older version of the Kakadu binaries. The documentation says that it needs at least version 7.0, but I'm not sure what exactly changed. As far as I can tell (in my relatively simple testing so far) everything seems to be working properly with version 6.

-Michael

### DIFF
--- a/doc/redhat-install.md
+++ b/doc/redhat-install.md
@@ -1,0 +1,217 @@
+Install on Red Hat (CentOS)
+===========================
+
+Installing Loris on Red Hat Enterprise Linux 6 (or the equivalent CentOS
+system) is tricky, since the bundled Python is not the right version and none
+of the `yum` packages are the latest versions. This guide is an attempt to
+guide you through the install process.
+
+Everything here needs to be done as root, so either open a root shell now or
+precede all instructions here with `sudo`. It also assumes you have `yum`,
+`wget`, and `git` in your path already (you may need to install git).
+
+Install Python 2.7
+------------------
+
+The first thing to do is to install Python 2.7 (RHEL 6 ships with 2.6). To do
+that, first you have to install the development tools (a C compiler and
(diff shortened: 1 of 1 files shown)


jpstroop:
Thanks a lot for this! I'd installed Loris on CentOS 5.5 early on in the project. I didn't have nearly this much trouble (had to get a newer Python, but that's about all). I'm guessing a lot has changed, and others have asked for RedHat/CentOS instructions recently, but we've since moved nearly all our machines to Ubuntu and I haven't had an easy way to see what the differences might be.

Regarding kakadu 6 vs 7, version >7 allows for other color profiles besides sRGB with JP2s. When this is the case, [Loris extract the color profile from the image header](https://github.com/pulibrary/loris/blob/development/loris/img_info.py#L198-L205) and [map it to sRGB](https://github.com/pulibrary/loris/blob/development/loris/transforms.py#L333-L335) for web safety (and hopefully, the best fidelity possible). To test this, have a look at [this image](https://github.com/pulibrary/loris/blob/development/tests/img/47102787.jp2) on your instance. It should [look like this](http://libimages.princeton.edu/loris/47102787.jp2/full/full/0/native.jpg) but chances are the names of the colors will be the wrong color.

This [feature can be disabled](https://github.com/pulibrary/loris/blob/development/etc/loris.conf#L66) and I'll make an note in the instructions you submitted explaing as much.

Thanks again!

jpstroop:
@mmcclimon, based on the info above, how does [this](https://github.com/pulibrary/loris/blob/development/doc/redhat-install.md#install-image-libraries) look to you?

mmcclimon:
Beautiful...thank you!

PR #124, opened 2014-09-15
Migrate docker container to passenger based container. Add documentation.

eliotjordan:


jpstroop:
Do we really need a copy of OpenSeadragon in the repo? If so, is there a way we could do it as a submodule?

eliotjordan:
Nope. I was trying to cram too much into that README. Those steps really belong somewhere else.

jpstroop:
OK. I'm definitely not against the documentation being in the repo (see the ticket I put in this AM for some ideas about reorganization suggestions). I think the Docker approach has loads of promise. Let me know if I can help.

PR #137, opened 2014-11-14
Initial stab at SimpleHTTPResolver Functionality

scande3:
This pull request adds a SimpleHTTPResolver with various common config
options. These options are as follows:

```
The config dictionary MUST contain
 * `cache_root`, which is the absolute path to the directory where
    source images should be cached.

The config dictionary MAY contain
 * `source_prefix`, the url up to the identifier.
 * `source_suffix`, the url after the identifier (if applicable).
 * `default_format`, the format of images (will use content-type
    of response if not specified).
 * `head_resolvable` with value True, whether to make HEAD requests
    to verify object existence (don't set if using Fedora Commons
    prior to 3.8).
 * `uri_resolvable` with value True, allows one to use full uri's
    to resolve to an image.
 * `user`, the username to make the HTTP request as if needed.
 * `pw`, the password to make the HTTP request as if needed.
```

Additionally, there is a new clean script titled
"loris-http_cache_clean.sh" in the bin directory with more granular
directory options.

Lastly: it now has a dependency of having the requests library installed
(pip install requests).

Known issues:
1. It still does a base bare identifier call that usually will return
   'false' first before trying more common routing cases.
2. The code to determine an identifier will strip out a slash in
   "http://". I currently compensate for this flaw but should bbe fixed
   in webapp.py's _dissect_uri.
3. Unit test coverage is sparse and needs to be expanded for more
   cases. Additionally, things need to be more properly mocked out
   as using a live url for the resolver at the moment.
4. The "links" and "info" directories are a mess regardless of
   resolver due to all entries not being subdivided. Not sure if this
   matters. In addition, it will write ":" if part of the identifier
   (not allowed on some file systems) along with symlinks being
   unsupported in some drive storage cases.

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -4,13 +4,17 @@
 ================================================
 """
 from logging import getLogger
-import loris_exception
+from loris_exception import ResolverException
 from os.path import join, exists, isfile
-from urllib import unquote
-from shutil import copy
 from os import makedirs
 from os.path import dirname
-from loris_exception import ResolverException
+from shutil import copy
+from urllib import unquote, quote_plus
+
+import constants
+import hashlib
+import glob
+import requests
 
 logger = getLogger(__name__)
 
@@ -30,7 +34,7 @@ class _AbstractResolver(object):
         Returns:
             bool
         """
-        e = self.__class__.__name__
+        cn = self.__class__.__name__
         raise NotImplementedError('is_resolvable() not implemented for %s' % (cn,))
 
     def resolve(self, ident):
@@ -47,7 +51,7 @@ class _AbstractResolver(object):
         Raises:
             ResolverException when something goes wrong...
         """
-        e = self.__class__.__name__
+        cn = self.__class__.__name__
         raise NotImplementedError('resolve() not implemented for %s' % (cn,))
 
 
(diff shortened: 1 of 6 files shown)


loris/resolver.py
```diff
@@ -85,6 +89,218 @@ def resolve(self, ident):
...
+    def __init__(self, config):
+        super(SimpleHTTPResolver, self).__init__(config)
+
+        # There may be a better way to do this...
+        if 'cache_root' in self.config:
+            self.cache_root = self.config['cache_root']
+        else:
+            message = 'Configuration incomplete and cannot resolve. Missing setting cache_root.'
+            logger.error(message)
+            raise ResolverException(404, message)
```
jpstroop: Should be a 500, no? Could use this:
https://github.com/pulibrary/loris/blob/development/loris/webapp.py#L205-L210

loris/resolver.py
```diff
@@ -85,6 +89,218 @@ def resolve(self, ident):
...
+     * `head_resolvable` with value True, whether to make HEAD requests to verify object existence (don't set if using
+        Fedora Commons prior to 3.8).
+     * `uri_resolvable` with value True, allows one to use full uri's to resolve to an image.
+     * `user`, the username to make the HTTP request as.
+     * `pw`, the password to make the HTTP request as.
+    '''
+    def __init__(self, config):
+        super(SimpleHTTPResolver, self).__init__(config)
+
+        # There may be a better way to do this...
```
jpstroop: You could do `self.__dict__ = config`, but that's always felt a little too magical to me. And you'd still wind up needing need to check what's defined.

loris/resolver.py
```diff
@@ -85,6 +89,218 @@ def resolve(self, ident):
...
+        if 'cache_root' in self.config:
+            self.cache_root = self.config['cache_root']
+        else:
+            message = 'Configuration incomplete and cannot resolve. Missing setting cache_root.'
+            logger.error(message)
+            raise ResolverException(404, message)
+
+        if 'source_prefix' in self.config:
```
jpstroop: For all of these if/elses that default to `''` or `None` I think you could do, e.g.

``` python
self.source_prefix = self.config.get('source_prefix', '')
```

where the default value is the second param passed to `#get`.

loris/resolver.py
```diff
@@ -103,7 +319,7 @@ def __init__(self, config):
 
     def is_resolvable(self, ident):
         ident = unquote(ident)
-        fp = join(self.cache_root, ident)
+        fp = join(self.source_root, ident)
```
jpstroop, line 312: Thanks!

jpstroop:
This is looking good! I will try to get a chance to run it over the weekend, but may not get to it until early next week.

Also (I may figure it out when I run it locally myself), but can you explain what you mean by:

> The "links" and "info" directories are a mess regardless of resolver due to all entries not being subdivided.

? 

Some other comments inline.

Best,
Js

jpstroop:
One other thing...re: testing. I think that as long as you're unit testing the individual methods you've introduced I'm OK with you not actually trying to do a GET that actually resolves to an image. At that point, arguably, you're really just testing that requests `get` method and that you can write the output to a file system.

scande3:
@jpstroop - can make the commented changes by the end of tonight and update this pull request with them for when you are able to try this. :)

As for the "links" and "info" directory comment, in a little over a week, we were up 19364 objects cached. These are then stored as a bunch of "commonwealth:<id>" in those directories. Running a simple "ls" in such directories then has a bunch to output to the console. :) Organizing them in subfolders to prevent any single directory from becoming so large likely isn't a functional concern though.

scande3:
@jpstroop - Got distracted over the weekend. >< Will have changes up by 11:00 AM EST. Thanks!

jpstroop:
No rush here!

scande3:
@jpstroop - Updated. Notably: there is now a requirement on mock and responses added. (May work without mock but unsure and cannot easily test since installed on my machine). This allows me to now prevent any HTTP calls using a very simple hexidecimal string for a tif (there may be a simpler valid string in another format...). I do much more granular testing on the format options.

(Can add documentation to this pull request or to another one if this all looks good).

jpstroop:
Just a quick thought about cache structure. Would turning, e.g. something like this:

```
iiif.digitalcommonwealth.org/commonwealth:3b591d48c
```

into:

```
org/digitalcommonwealth/iiif/commonwealth/3b/59/1d/48/c.jp2
```

work/help? Not saying you should do it now, just an idea.

scande3:
@jpstroop - That won't affect what is inside of "links" and "info" at all since that is handled outside of the resolver itself.  Beyond the directories being thousands of "commonwealth:*", I had originally tried to put all of that on our NAS as well. Sadly, our NAS doesn't support ":" in filenames nor did it support creating symlinks. Just mentioned it in my pull issues as something to perhaps make more optimal in the future (if possible) but they don't cause a current functional issue.

But regarding the idea: An earlier implementation I was trying did do something similar to what you have posted. I ended up scrapping it for now since I didn't know what types of identifiers exist outside for Fedora Commons / URIs and didn't want to risk something breaking for an unforeseen identifier case. Can make it that way if that is a storage preference over the current way it is functioning?

jpstroop:
No problem, I'm not bothered for now; just an idea.

PR #138, opened 2014-11-17
Documentation update for SimpleHTTPResolver

scande3:
Update the documentation for the new resolver. In addition, fix any
errors I am aware of in it. And add functionality to move the Simple
HTTP resolver clean script to where the other clean script is place.

These are the instructions I had to use for it to run under Ubuntu
14.04.

Let me know if you have anything that should be changed. Thanks!

### DIFF
--- a/doc/apache.md
+++ b/doc/apache.md
@@ -28,7 +28,7 @@ ExpiresDefault "access plus 5184000 seconds"
 AllowEncodedSlashes On
 
 WSGIDaemonProcess loris user=loris group=loris processes=10 threads=15 maximum-requests=10000
-WSGIScriptAlias /loris /var/www/loris/loris.wsgi
+WSGIScriptAlias /loris /var/www/loris/loris2.wsgi
 WSGIProcessGroup loris
 ```
 
--- a/doc/cache_maintenance.md
+++ b/doc/cache_maintenance.md
@@ -1,9 +1,17 @@
 Caching
 =======
 
+### `SimpleFSResolver`
+
 There is a Bash script at `bin/loris-cache_clean.sh` that makes heavy use of `find` command line utility to turn the filesystem cache into a simple LRU-style cache. Have a look at it and set the constants near the top; it is intended to be deployed as a cron job.
 
 After you run `setup.py install` this script will be a `/usr/local/bin/loris-cache_clean.sh`.
+
+### `SimpleHTTPResolver`
+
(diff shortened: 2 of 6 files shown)


scande3:
(And just fixed a bug with the new cache clean movement in setup.py as I just installed it fresh on our server. I'll do some testing on the http clean script over the next few days as admittedly haven't tested that much yet).

jpstroop:
Awesome, thanks!!

rlskoeser:
Shouldn't the python deps documented in the dependencies file be added to the setup.py requirements so they can be installed automatically?  

And this is more a question for @jpstroop, but something I noticed looking at the changes - is there any reason you're not using the scripts option in setup.py for installing utilities like the cache cleanup scripts? Are you using data_files because you need more control over where they go?

jpstroop:
@rlskoeser If you have any advice on how to do the deployment more cleanly, I'm totally happy to hear it. I've never felt that comfortable with it as it seems to be a constantly moving target `setuptools`, `distutils`; I can't keep it all straight! (I'm not trying to be snippy, I really hate it! :smile: )

There is a specific chicken-egg problem in that `configobj` is required by setup.py, which means that setup.py will always fail without it. So instead we generally do:

```
pip install -r requirements.txt
```

first, and then

```
python setup.py install
```

I've always hated that, but shudder at the idea of an install script that requires a gazillion flags for options. I try to handle the chicken-egg problem [here](https://github.com/pulibrary/loris/blob/development/setup.py#L20-L30), but it's far from ideal. 

Re: `scripts` vs `data_files`, I believe it was about controlling where they wind up, but it's been a looong time. 

I'd accept any PR that made this easier! :wink:

rlskoeser:
I noticed the configobj dependency; that is a little troubling.  Why is it needed before setup can run?  Is the config file supposed to exist before loris is even installed?

It does seem to me that any dependencies _besides_ configobj should be handled in the setup script so at least there is only one special case (and maybe we can figure out how to handle that too).

FYI, it occurs to me that hard-coding the script install paths likely means that loris probably wouldn't install very well inside a virtualenv - we use virtualenv all the time to isolate different projects, and it can also be useful in production if you have different projects running on the same box.

I tried to look at your setup script before and it seemed pretty complicated.  I guess it is doing a lot of work to get everything set up the way it needs to be.  I'll try to think about it and see if I might be able to contribute (no promises, I don't have much time to work on this at the moment).  I know when we first started looking at this the sysadmin I was working with was not impressed with the installation and setup process, so maybe I can enlist him to help me streamline it.

PR #139, opened 2014-11-19
Adds a corrupt jp2 for testing and handles it as best as it can

jpstroop:
@schwemmer 

Could you have a look at this branch? Because the shell-out to Kakadu is in a subprocess, my options are fairly limited. However, I wasn't cleaning up in the event that _something_ goes wrong during the transformation,  and I've [added a generic exception handling block](https://github.com/pulibrary/loris/compare/handle_corrupt_jp2?expand=1#diff-05d372ac8868d6959ce12ba84449ee34R423) per your suggestion. Additionally, you should see, e.g.:

```
2014-11-19 15:01:04,986 (transforms) [ERROR]: Kakadu Core Error:
2014-11-19 15:01:04,986 (transforms) [ERROR]: Missing or out-of-sequence tile-parts for tile number 38 in code-stream!
```

in your logs and something along the lines of:

```
Server Side Error: image was incomplete 

(/home/jstroop/workspace/loris/tests/img/corrupt.jp2). (500)
```

On your screen. Hopefully this keeps you machine from being overwhelmed--I haven't been able to reproduce it.

### DIFF
--- a/loris/transforms.py
+++ b/loris/transforms.py
@@ -393,31 +393,38 @@ class KakaduJP2Transformer(_AbstractJP2Transformer):
 
         logger.debug('Calling: %s' % (kdu_cmd,))
 
-        # Start the kdu shellout. Blocks until the pipe is empty
-        kdu_expand_proc = subprocess.Popen(kdu_cmd, shell=True, bufsize=-1, 
-            stderr=subprocess.PIPE, env=self.env)
+        try:
+            # Start the kdu shellout. Blocks until the pipe is empty
+            kdu_expand_proc = subprocess.Popen(kdu_cmd, shell=True, bufsize=-1, 
+                stderr=subprocess.PIPE, env=self.env)
 
-        f = open(fifo_fp, 'rb')
-        logger.debug('Opened %s' % fifo_fp)
+            f = open(fifo_fp, 'rb')
+            logger.debug('Opened %s' % fifo_fp)
 
-        # read from the named pipe
-        p = Parser()
-        while True:
-            s = f.read(1024)
-            if not s:
-                break
-            p.feed(s)
-        im = p.close() # a PIL.Image
+             # read from the named pipe
+            p = Parser()
+            while True:
+                s = f.read(1024)
+                if not s:
+                    break
+                p.feed(s)
+            im = p.close() # a PIL.Image
 
-        # finish kdu
-        kdu_exit = kdu_expand_proc.wait()
-        if kdu_exit != 0:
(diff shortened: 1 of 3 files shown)


schwemmer:
Awesome, this totally fixes it! Works great. Not only the memory leaks and CPU load at 100% are gone, also the images display normally now, even the corrupted ones (maybe because I am not using tiles). I used the same corrupted image I sent you for testing and lots of other ones. 

The only thing I can not see anywhere in any log is the error message that "there was a problem with the source file". But again, this might be because I am not using tiles.

jpstroop:
You should be seeing

```
2014-11-19 15:01:04,986 (transforms) [ERROR]: Kakadu Core Error:
2014-11-19 15:01:04,986 (transforms) [ERROR]: Missing or out-of-sequence tile-parts for tile number 38 in code-stream!
```

in your logs. That's not showing up?

I'll merge this now.

schwemmer:
Yes, you're right, sorry about that, I just noticed it as well. And indeed, I can see that in my log. Only the error on the screen is not there, instead I see the actual image which is nothing to complain about ;-)

jpstroop:
Wait, WHAT? The corrupted image is working? That's bizarre! Can you clear your cache and try again? I wasn't even able to get thumbnails before

schwemmer:
Ok, let me try to clean the cache. I found it bizarre too. But I am pretty sure it is not in the cache, I just added it to a test manuscript under a different name and it worked right away.

schwemmer:
I can confirm that the corrupted image is definitely working. These are the steps I did to confirm:
- Deleted entire cache directory
- Verified that cache directory is non-existent
- Disabled caching in etc/loris/loris2.conf
- Restarted Apache
- Loaded corrupted image in viewer
- Loaded corrupted image in viewer in different browser after deleting browser cache to rule out client-side caching
- Image loads normally
- /var/log/loris/loris.log says: `2014-11-20 00:16:55,510 (loris.transforms) [ERROR]: Kakadu Core Error:
  2014-11-20 00:16:55,510 (loris.transforms) [ERROR]: Missing or out-of-sequence tile-parts for tile number 38 in code-stream!`

schwemmer:
First I thought that it is maybe working because I am not using tiles. But since you say you weren't even able to get the thumbnail, it must be something else. Maybe because I am using Kakadu 7.4 and Loris comes with Kakadu 7.2 out of the box? Just a wild guess.

jpstroop:
Must be the difference in Kakadus....I'll to upgrade next week and see if it makes a difference on my end.

Issue #98, opened 2014-07-21
Implement an HTTP Resolver


jpstroop:
A few questions here:
- Does the whole URI go into the identifier slot (escaped, of course)?:
  
  ``` python
  >>> from urllib import quote_plus
  >>> quote_plus("http://iiif-test.stanford.edu/67352ccc-d1b0-11e1-89ae-279075081939.jp2")
  'http%3A%2F%2Fiiif-test.stanford.edu%2F67352ccc-d1b0-11e1-89ae-279075081939.jp2'
  ```
- Or just some part of the URI that gets plugged in (config could supply a string template):
  
  ``` ini
  [resolver]
  impl=HTTPResolver
  template=http://example.edu/images/%s
  ```
  
  so that the identifier slot could just be a (more succinct, tidier) portion of the URI. Option 1 (above) and this may not be mutually exclusive, i.e. if there's a template specified, use it, if not, don't. 
  
  Could there be more than one template? Something like:
  
  ``` ini
  [resolver]
  impl=HTTPResolver
  templates=a,b
  a=http://example.edu/images/%s
  b=http://example.edu/images-elsewhere/%s
  ```
  
  and then the identifier would look like: `a%3Amy_id` (`a:my_id`), `b%3Amy_id` (`b:my_id`).
- How should the format be determined? Content-Type from the response seems obvious, as long as it's reliable. To what extent should this be configurable--we could fall back to the suffix on the identifier, but it would be better to have it stated explicitly so that the resolver can be configured once, rather than do a bunch of string checking on each call to `resolve`:
  
  ``` ini
  [resolver]
  impl=HTTPResolver
  format_source=suffix ; or content-type
  ```

jpstroop:
These are the thoughts about an HTTP resolver that have been floating around in my head, whether @lfarrell does it or someone else. @rlskoeser any thoughts about this?

ebenenglish:
Just chiming in as we're interested in this issue as well...

Identifiers:
I think to be as flexible as possible, you'd want to pursue the "if there's a template specified, use it, if not, don't" option. For our implementation, we'd probably specify the URL-root template in the config so the URLs that get constructed for Loris can be simpler/cleaner/shorter by just using the image identifier. Multiple templates would perhaps be overkill if you have the ability to specify a full URL in the identifier?

Formats:
[...]

Issue #95, opened 2014-07-19
Image format identification


jpstroop:
See [this comment](https://github.com/pulibrary/loris/issues/58#issuecomment-49485717). It's true, and this is a fairly typical convention for image processing libraries (as I wrote [here](http://iiif.io/api/image/2.0/change-log.html#dropped-support-for-content-negotiation-and-default-image-format) when IIIF dropped support for conneg) and I don't feel that bad about it.

**However**, with the current resolvers, if your local file is named for just some opaque number or string, you're hosed. One option would be to use something like [ahupp/python-magic](https://github.com/ahupp/python-magic) (basically a wrapper for the Unix file utility). So if a jp2 came onto the file system and it was named "761390", you could do the Python equivalent of:

``` bash
$ file --mime-type 761390
761390: image/jp2
```

with

``` python
>>> import magic
>>> magic.from_file("local_cache/761390", mime=True)
'image/jp2'
```

And the mime-type could then be mapped to, e.g. `jp2` or whatever is appropriate.

I don't think I'd want to run that for every request (though maybe it would be OK), so at least an in-memory hash (`OrderedDict`?) mapping IDs to formats would probably be a good idea.

Not saying I'll add this functionality any time soon, but it's an option, and needed to be logged.

jpstroop:
(In reference to https://github.com/pulibrary/loris/issues/98#issuecomment-49982002)

@eocarragain Thanks! Ruven and I had a brief discussion about this last winter--nice to see he has a solution. His approach is probably a lot more efficient and makes a lot of sense: using a big library like I proposed in #95 is probably overkill when we could just as easily sniff at the first few bytes of the file--we only have a handful of signatures to worry about, not thousands!

Another approach, if we did want the nice tidy API that [ahupp/python-magic](https://github.com/ahupp/python-magic/) provides would be to supply a shorter magic file, which is an option:

https://github.com/ahupp/python-magic/blob/master/magic.py#L37-L53

and also limit how much we read from the file:

https://github.com/ahupp/python-magic/blob/master/magic.py#L13-L15

Just to log it, there's some info about how to do that here: http://stackoverflow.com/a/7236262/714478

This is starting to feel more compelling....

PR #155, opened 2015-04-29
Add extension normalizing resolver.

hading:
Perhaps someone else will find this useful.

We have a lot of files with different extensions (e.g. for tifs, .tif, .tiff, .TIF, .TIFF). They're preservation files, so we can't normalize the extensions on the file system.

This derivative of the SimpleFSResolver first downcases the file extension and then does a lookup configured in the loris configuration file to try to convert these to a standard loris file type.

It'd be ideal to guard against the absence of the extension map section in the config file, but Python isn't really a language I know well and I don't know an idiomatic way to do that here.

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -89,6 +89,28 @@ class SimpleFSResolver(_AbstractResolver):
 
         return (fp, format)
 
+#To use this the resolver stanza of the config will have to have both the
+#src_img_root as required by the SimpleFSResolver and also an
+#[[extension_map]], which will be a hash mapping found extensions to the
+#extensions that loris wants to see, e.g.
+# [resolver]
+# impl = 'loris.resolver.ExtensionNormalizingFSResolver'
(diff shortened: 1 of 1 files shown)


jpstroop:
This is a great patch, thanks! At some point #95 would hopefully in alleviate the need for any file extensions in the id, but for now this is definitely useful.

hading:
No problem. 

We're toying with the idea of some other ways to identify the file type (we may have identified it via 'file' or FITS, and at some point may be using http where it will be available in the Content-Type header). As much flexibility as is available around this is good.

At some point we may try doing a resolver that can do both files and http and multiple paths/prefixes as well. I think this would also be fairly straightforward and if we get to the point that we need it and nothing to do that has appeared yet I'll drop that your way too.

jpstroop:
Have a look at #98...it's a long issue, so maybe start [here](https://github.com/loris-imageserver/loris/issues/98#issuecomment-50018589). I sketched this out a while ago, and think it would be great to have as a sort of universal resolver.

PR #159, opened 2015-06-01
adds missing favicon

eliotjordan:
setup.py fails without favicon.ico

```
Changed ownership of /var/www/loris2/icons to loris:loris
error: [Errno 2] No such file or directory: '/opt/loris/www/icons/favicon.ico'
```

This adds it back using loris-circle.png. The other option is to rewrite setup.py and remove references to favicon. Which do you prefer @jpstroop?


jpstroop:
Hmm...that's interesting...surprised that hadn't come up sooner.

PR #163, opened 2015-06-29
Tweak _web_request_url for uri_resolvable case in SimpleHTTPResolver

rlskoeser:
I was trying out the current release of loris and I couldn't get it to work using full urls as the identifier, and I finally tracked it down to this section.

There's a note in the code that it's working around a problem where urls were coming in with only a single slash after http/https - that's not what's happening for me (and I tested this in a couple of different environments)--  and the fix was generating invalid urls with three slashes.  I've adjusted the url cleanup so it will only add two slashes - although technically my fix would clean up and work with any number of slashes after http/https, which may not be completely ideal either.

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -208,9 +208,11 @@ class SimpleHTTPResolver(_AbstractResolver):
     @staticmethod
     def _web_request_url(ident, is_uri_resolvable, prefix, suffix):
         if (ident[0:6] == 'http:/' or ident[0:7] == 'https:/') and is_uri_resolvable:
-            #ident is http request with no prefix or suffix specified
-            #For some reason, identifier is http:/<url> or https:/<url>? Hack to correct.
-            return ident[0:ident.find('/')] + '/' + ident[ident.find('/'):len(ident)]
+            # ident is http request with no prefix or suffix specified
+            # For some reason, identifier is http:/<url> or https:/<url>? 
+            # Hack to correct without breaking valid urls.
+            first_slash = ident.find('/')
+            return '%s//%s' % (ident[:first_slash], ident[first_slash:].lstrip('/'))
         else:
             return prefix + ident + suffix
 


jpstroop:
Ping @scande3

scande3:
This logic makes sense to me.

(Side note: we are still using the version of Loris from 2014ish. There is plan sometime over the next two weeks to update that and to fix documentation that is no longer up to date. I'll likely also add some documentation of Ruby code to invalidate the cache on a particular item as well. Essentially if one replaces an image using Hydra, how does one then remove the Loris cached version of the outdated image to not have to wait for some cron job to eventually clean that up).

scande3:
@jpstroop - For you to merge as I don't have any permissions on this repo. :+1:

rlskoeser:
Thanks for the quick feedback and merge!

@jpstroop any idea when this will make it into a release?

jpstroop:
There's a glimmer of hope that I might be able to spend a week or so doing some refactoring and other work on Loris in August.

If that's the case, one of my goals would be to get rid of [one](https://github.com/loris-imageserver/loris/blob/development/loris/webapp.py#L317), if not [both](https://github.com/loris-imageserver/loris/blob/development/loris/webapp.py#L263), calls to `is_resolvable()` in webapp.py. The former call is a hack that is currently the only way the `route()` method can distinguish between a 400 and a 404, and it's now been [confirmed by the horse's mouth](http://stackoverflow.com/q/30981696) that there is no way to distinguish between `%2F` and `/` in WSGI applications, so the solution is likely to be a breaking change that requires a configurable replacement character for `/`s in identifiers in the `SimpleFSResolver`. (I see this as a breaking change because it's the default implementation, and I know that there's a fair number of folks using it).

TMI, I realize, but the point is that I'd rather not cut a 2.1.0 release shortly followed by 3.0.0. So let's say that, one way or another, that there will be a release by the end of August, and if I'm lucky it'll be 3.0.0, and if not, then 2.1.0. Does that sound OK?

rlskoeser:
Thanks for the info, that's helpful - and certainly makes sense from your perspective.  

I'm hoping to get my sysadmins looking into setting up loris soon - now that you've cleaned up the deploy, hopefully it will look more appealing to them than continuing to run djatoka.  If I can get some traction, we may just have to run a patched version for a while (assuming anything happens quickly, although it might not happen as fast as I'd like) - but I think they'll be more comfortable knowing we can switch to the official release whenever the next version comes out.

PR #173, opened 2015-07-06
Make apt-get command cut-and-paste-able

rlskoeser:
This is a tiny edit, but it makes the install a little bit easier because the multi-line apt-get command can be cut and paste as a single line.


jpstroop:
:+1:

PR #183, opened 2015-10-19
Fix TemplateHttpResolver warning for no templates configured

rlskoeser:
I was running this locally and noticed the test for warning that no templates configured is wrong. Whoops!


jpstroop:
Hah! Oops :wink:

PR #186, opened 2015-10-27
Avoid and/or close streaming requests under SimpleHTTPResolver

ghukill:
We had an interesting scenario with Mirador and the `SimpleHTTPResolver` that was resulting in thumbnails (and occasionally the main image) to load incompletely, and then cache that incomplete sourcing, resulting in images with gray chunks.  The behavior seemed to happen when Loris would get a bunch of concurrent requests for images while using an HTTP resolver, which led us to think it was related to communications between Loris and the HTTP source images.  

Noticed two things:

1) The `is_resolvable` method from the `SimpleHTTPResolver` does not close the streaming request when checking the HTTP status code (https://github.com/loris-imageserver/loris/blob/development/loris/resolver.py#L180-L192).  This doesn't appear to be what caused the images to not load entirely, but it's tangentially related, and it seems like it couldn't hurt to close this request once the code has been read?  

2)  The actual cause appears to be under the `resolve` method of the `SimpleHTTPResolver`, which also opens a streaming request, that is eventually chunked to file on disk.  For reasons I'm still a little unclear, I think requests from Mirador that try and retrieve multiple images at once will cause that connection to close before the chunking can finish.  Adding a small delay in that chunking process - like writing to the log - results in thumbnails and images that are mostly gray, as if they didn't get close to finishing.  But by changing that `stream = True` to `stream = False`, I have not been able to recreate the problem with images sourced via HTTP not fully loading; dozens of multiple requests load correctly.

Thought perhaps our slightly underpowered development environment, coupled with a heavy concurrent request load, revealed this behavior, and that these conditions wouldn't manifest on a more powerful setup.  But it doesn't seem to slow the loading of multiple images much at all, and has seemed to fix the problem!  Thought it was worth mentioning and submitting here.  Thanks.

_Note_: This pull request includes changes from #185, as both modify behavior in the `loris/resolver.py` and were a bit intertwined.

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -10,6 +10,7 @@ from os import makedirs
 from os.path import dirname
 from shutil import copy
 from urllib import unquote, quote_plus
+from contextlib import closing
 
 import constants
 import hashlib
@@ -179,17 +180,19 @@ class SimpleHTTPResolver(_AbstractResolver):
 
             if self.head_resolvable:
                 try:
-                    response = requests.head(fp, **self.request_options())
-                except requests.exceptions.MissingSchema:
-                    return False
-            else:
-                try:
-                    response = requests.get(fp, stream = True, **self.request_options())
+                    with closing(requests.head(fp, **self.request_options())) as response:
+                        if response.status_code is 200:
+                            return True
                 except requests.exceptions.MissingSchema:
                     return False
 
-            if response.status_code is 200:
-                return True
+            else:
+                try:
+                    with closing(requests.get(fp, stream=True, **self.request_options())) as response:
+                        if response.status_code is 200:
+                            return True
+                except requests.exceptions.MissingSchema:
+                    return False
 
         return False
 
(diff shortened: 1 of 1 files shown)


loris/resolver.py
```diff
@@ -177,20 +178,10 @@ def is_resolvable(self, ident):
         else:
             fp = self._web_request_url(ident)
 
-            if self.head_resolvable:
```
jpstroop, line 180: @ghukill: Why does this check go away?

ghukill, line 180: Looking at it now, I'm not sure why.  My best guess is that I got tangled up in the circular symlinks (https://github.com/loris-imageserver/loris/issues) and this was a way to short-circuit while testing.  Whatever the case, and my apologies for letting that persist into the pull request, I think you're right that it should stay.

If you want to go with the `with closing(...` syntax, that'd mean rewriting the requests from lines 181 - 193 of the original code to use that syntax.

What's your preference for doing this?  Would it be easier if I created a new pull request?

jpstroop, line 180: I really like the `with closing(` statement, but I'm pretty sure that other users (@scande3 @rlskoeser ?) need that head request check. If you can update lines 181 - 193 to use a similar syntax, I think that should be the plan.

No need for a new PR unless you want to--I'm fine with you just updating this branch and squashing the commits.

ghukill, line 180: Sounds good!  Will do.

ghukill:
My commit squashing foo is debatable, but I'm hoping the `HEAD` check is reinstated correctly, and the `with closing(` statements came through alright. Thanks for your patience.

jpstroop:
I see one commit with one parent, so it seems like your squash-fu is fine! Thanks for _your_ patience...I'm sorry I wasn't able to get to this sooner.

Issue #187, opened 2015-11-10
500 error / circular symlinks


rlskoeser:
I'm occasionally seeing 500 errors with circular symlinks for images that at other times have worked fine.

Example output from loris:

```
Server Side Error: [Errno 40] Too many levels of symbolic links: u'/u/data/loris/cache/qasfedora:readux:px393/full/!100,100/0/default.png' 

This is likely a permissions problem, though it's
possible that there was a problem with the source file
(/u/data/loris/cache/qasfedora/readux/16/7c3/8ba/44b/dd0/f1e/b15/b98/f5b/b8b/d30/loris_cache.jp2). (500)
```

The original cache file looks fine, but the resized version appears to be a symlink to itself:

```
ls -l '/u/data/loris/cache/qasfedora:readux:px393/full/!100,100/0/default.png'
lrwxrwxrwx 1 loris apache 70 Nov 10 10:20 /u/data/loris/cache/qasfedora:readux:px393/full/!100,100/0/default.png -> /u/data/loris/cache/qasfedora:readux:px393/full/!100,100/0/default.png
```

I may be able to look into this if someone can provide some suggestions of where in the code to start looking or some ideas on what the problem might be.  As I said, it's sporadic because these kinds of images display fine most of the time.

jpstroop:
I can't look at the moment, but it's either the code in the image cache or
the code in the webapp that makes the image that makes these symlinks. The
idea is to store the image in the cache at a path based on the canonical
request URI, and then link to it from a path based on the actual request.

My guess is that there's a bug in the condition that checks whether or not
the request was made using the canonical syntax.

It's a minor performance enhancement, and I wouldn't be against removing it
altogether. It's not the first time it's caused trouble.

-Js

Sent via mobile. Please excuse typos, brevity, etc.

rlskoeser:
As you'll see, I've got a pull request with a preliminary fix to prevent creating circular symlinks, but I haven't figured out yet what could cause a canonical cache path to be generated incorrectly.  I'm guessing maybe it's in the parameters code; is it possible that the size parameter could fail to get the image size under certain circumstances and thus not be able to calculate the normalized form correctly? 

[...]

PR #190, opened 2015-11-11
Prevent creation of circular symlinks

rlskoeser:
## Preliminary fix for #187

This update doesn't address whatever issue is resulting in circular symlinks being created, but it adds a check to prevent actually creating them.  Based on the logic in ImageCache._link, I think in the cases where this is cropping up, loris is actually _removing_ a valid image file in order to create the broken circular symlink.  So, this fix will prevent that at least as a stop-gap until someone can determine what's causing it to happen.

### DIFF
--- a/loris/img.py
+++ b/loris/img.py
@@ -216,6 +216,11 @@ class ImageCache(dict):
 
 	@staticmethod
 	def _link(to,frum):
+		# under certain circumstances, files are generating circular
+		# symlinks; don't do that
+		if to == frum:
+			logger.warn('Circular symlink requested from %s to %s; not creating symlink' % (frum, to))
+			return
 		link_dp = path.dirname(frum)
 		if not path.exists(link_dp):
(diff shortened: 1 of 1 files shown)


jpstroop:
@rlskoeser spending some time with this this morning; as discussed. I may just wind up taking out the symlinking feature. Will be back in touch.

rlskoeser:
@jpstroop glad you have some time to look into it! I haven't had much time to dig in, and was concerned that there was additional logic somewhere beyond the section you pointed me to, since it looks like the non-canonical path is already in use as the fp parameter set when the code gets to this point:
https://github.com/loris-imageserver/loris/blob/development/loris/img.py#L227

Looking forward to seeing how you fix it.

jpstroop:
It's going to take a little time. My instinct is still to remove the symlinking altogether, but I noticed that there's no test coverage for the cache, so starting there before I change anything.

PR #191, opened 2015-11-23
Symlinks were going the wrong direction

### LINKED ISSUE #187: 500 error / circular symlinks

jpstroop:
@rlskoeser The symlinks were going the wrong way--i.e. from the canonical path TO the non-canonical version. Although I can't exactly imagine how this would result in circular links, I guess it's possible. [I added tests](https://github.com/loris-imageserver/loris/blob/remove_symlink_feature/tests/img_t.py#L22-L57) that were all failing until I made some very minor tweaks. It gets confusing very quickly, and I wouldn't mind a review if you have a moment.

If this doesn't solve your problem, It think we should remove the symlink stuff altogher (as I've been threatening)--it's a very minor savings at this point.

Closes #187.

### DIFF
--- a/loris/img.py
+++ b/loris/img.py
@@ -42,10 +42,12 @@ class ImageRequest(object):
 		info (ImageInfo):
 		is_canonical (bool):
 			True if this is a canonical path.
-		cache_path (str):
-			Relative path from the cache root, based on the original request values.
-		canonical_cache_path (str):
-			Relative path from the cache root, based on normalized values
+		as_path (str):
+			Useful as a relative path from a cache root. Based on the original
+			request values.
+		canonical_as_path (str):
+			Useful as a relative path from a cache root. Based on values
+			normalized to the canonical request syntax.
 		request_path
 			Path of the request for tacking on to the service host and creating
 			a URI based on the original request.
@@ -148,7 +150,8 @@ class ImageRequest(object):
 	@property
 	def canonical_request_path(self):
 		if self._canonical_request_path is None:
-			p = '/'.join((quote_plus(self.ident),
+			p = '/'.join((
+			    quote_plus(self.ident),
 				self.region_param.canonical_uri_value,
 				self.size_param.canonical_uri_value,
 				self.rotation_param.canonical_uri_value,
@@ -158,7 +161,7 @@ class ImageRequest(object):
 		return self._canonical_request_path
 
 	@property
-	def cache_path(self):
+	def as_path(self):
 		if self._cache_path is None:
 			p = path.join(self.ident,
 				self.region_value,
@@ -170,21 +173,21 @@ class ImageRequest(object):
 		return self._cache_path
 
 	@property
-	def canonical_cache_path(self):
+	def canonical_as_path(self):
 		if self._canonical_cache_path is None:
 			p = path.join(self.ident,
-					self.region_param.canonical_uri_value,
-					self.size_param.canonical_uri_value,
-					self.rotation_param.canonical_uri_value,
-					self.quality
-				)
+				self.region_param.canonical_uri_value,
+				self.size_param.canonical_uri_value,
+				self.rotation_param.canonical_uri_value,
+				self.quality
+			)
 			self._canonical_cache_path = '%s.%s' % (p, self.format)
 		return self._canonical_cache_path
 
 	@property
 	def is_canonical(self):
 		if self._is_canonical is None:
-			self._is_canonical = self.cache_path == self.canonical_cache_path
+			self._is_canonical = self.as_path == self.canonical_as_path
 		return self._is_canonical
 
 	@property
@@ -206,7 +209,7 @@ class ImageCache(dict):
 		self.cache_root = cache_root
 
 	def __contains__(self, image_request):
-		return path.exists(self.get_cache_path(image_request))
+		return path.exists(self.get_request_cache_path(image_request))
 
 	def __getitem__(self, image_request):
 		fp = self.get(image_request)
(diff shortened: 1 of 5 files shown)


loris/img.py
```diff
@@ -215,21 +215,21 @@ def __getitem__(self, image_request):
 		return fp
 
 	@staticmethod
-	def _link(to,frum):
+	def _link(to, frum):
```
rlskoeser: Perhaps renaming the variables to something like src and dest would make this a little clearer?  (target/link?) For some reason I find the to/from not intuitive.

loris/img.py
```diff
@@ -215,21 +215,21 @@ def __getitem__(self, image_request):
...
 		link_dp = path.dirname(frum)
 		if not path.exists(link_dp):
 			makedirs(link_dp)
 		if path.lexists(frum): # shouldn't be the case, but helps debugging
```
rlskoeser: I think this part might still be problematic, _if_ something goes wrong with calculating a path as I've been seeing in my loris instance.  Or, include my pull request to bail out if somehow and to and from values are the same.

loris/img.py
```diff
@@ -215,21 +215,21 @@ def __getitem__(self, image_request):
...
 			unlink(frum)
-		symlink(to,frum)
-		logger.debug('Made symlink from %s to %s' % (to,frum))
+		symlink(to, frum)
+		logger.debug('Made symlink from %s to %s' % (frum, to))
 
-	def __setitem__(self, image_request, fp):
-		# The Image (fp) already exists, this simply makes a symlink in the
-		# cache to from the canonical syntax to the actual request path
+	def __setitem__(self, image_request, canonical_fp):
```
rlskoeser, line 230: Perhaps not so relevant to these particular changes, but it's a little confusing to me that setting the item in the cache just creates the symlink; where is the code that actually saves the canonical version and how does it relate to whether or not caching is configured in the webapp.py logic that calls this?

rlskoeser:
Storing the image at the canonical path seems like an important fix.  I've looked over your changes and made a few comments, but I feel like I'm still missing something because I don't understand what saves the image at the canonical path if this change only affects the cache and symlink.  Do you need to change something elsewhere to make sure a requested image is always saved at the canonical path?

I haven't been able to figure out what in the code could be causing the circular symlinks, and it's harder to track down because it doesn't happen all the time.  As best as I can figure, it would mean that they were calculated correctly once (because they were not equal and the link code was triggered) and then calculated incorrectly (equal paths -> circular symlink).   I'm having trouble following how the SizeParameter calculates its canonical form, but it would have the actual image size to do so; is it possible it could fail to get the size under certain circumstances, and if so what would happen?

loris/img.py
jpstroop: I have the opposite problem (src, dest confuses me), but going with the descriptors used in the python docs would probably make sense: https://docs.python.org/2/library/os.html#os.symlink

loris/img.py
jpstroop: I'm happy to merge your PR. I wish I knew what was causing it to happen, but better safe :smile:

loris/img.py
jpstroop, line 230: Yes, I agree, this is the one place where my half-baked effort to implement the dict API breaks a little bit. The problem is that PIL and Kakadu need a path for the output image, and it doesn't make any sense to move the file (elsewhere to get it "in" to the cache)  once it's been made. The workaround is that the [webapp asks the cache for a path](https://github.com/loris-imageserver/loris/blob/development/loris/webapp.py#L588) based on the request up front, `__setitem__` is called later on when we know everything worked.

I'm more than happy to hear alternatives! Thanks for the review!

loris/img.py
rlskoeser: Agreed, src/dest isn't that clear either, since this isn't a move or copy.  +1 for using source and link_name

loris/img.py
rlskoeser, line 230: It would be helpful if you could add some documentation or comments to the ImageCache class to make it a bit clearer what it is used for and how it works.  That would make it easier if contributors need to do anything with that part of the code.

If I come up with alternatives, I'll certainly let you know.  And glad the review was helpful!

jpstroop:
OK, rebased and updated. Let me know what you think. I'll do a manual merge (I'm sure it won't merge automatically) on your PR once we feel like this is in a good place. I'm still skeptical about symlinks being worth the trouble...

rlskoeser:
The comment explaining the file and caching stuff is great, really helpful.

PR #192, opened 2015-11-23
Optional regex for CORS domains may be supplied

### LINKED ISSUE #181: documentation on how to set Access-Control-Allow-Origin
I'm looking for how to set Access-Control-Allow-Origin to something other than *. I see there are some commits making CORS-related changes. I expected to find documentation on how to do this here:
https://github.com/loris-imageserver/loris/blob/development/doc/configuration.md

Is there already an option to change this setting?

My use case for needing to change this is that until we can work out some policy issues around sharing images we may not be able to set the header to "Access-Control-Allow-Origin: *" so we would like to whitelist our domains.

jpstroop:
@jronallo work for you?

Closes #181 (maybe)

### DIFF
--- a/etc/loris2.conf
+++ b/etc/loris2.conf
@@ -27,6 +27,13 @@ run_as_group = 'loris'
 enable_caching = True
 redirect_canonical_image_request = False
 redirect_id_slash_to_info = True
+# cors_regex = ''
+# NOTE: If supplied, cors_regex is passed to re.search():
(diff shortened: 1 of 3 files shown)


jpstroop:
Merging. @jronallo, let me know if this works for you, and file another issue if not.

jronallo:
@jpstroop Thanks. Sorry I haven't gotten to this, but I'm going to cycle back to it at some point.

jpstroop:
:+1:

PR #194, opened 2015-11-25
Keep http and https info separate

### LINKED ISSUE #177: Make @id responsive to http and https?
I would like to support both http and https on my server. Is that possible to adjust loris to response @id based on incoming requests? Currently I always get http://... in the @id field no matter what protocol I use. Thanks.

jpstroop:
@khongtac2k, please let me know how this works out!
Closes #177


ta1986:
Just got a chance to test it with https. It works! Thanks much.

jpstroop:
Great!

PR #199, opened 2015-12-23
sizeAboveFull

gvasold:
I noticed that Loris has no restriction for size. This means that a user can force the server to interpolate an image to any size (eg. /1000000,/) which can be used for DoS attacks. So I added code to limit the size parameter. What i did:
- Introduced a new config parameter Loris.max_size_above_full. It expects a numeric value which restricts the maximum size to n percent of the original image size (width or height: not number of pixels). 100 means that the maximum size of the original image size must not be exceeded, 200 allows a maximum width or height which is twice the original width or height. I set the default value to 200 in Loris.**init**(). If max_size_above_full ist set to 0, no size restrictions are set, so Loris behaves like it did before.
- Made changes to get_image() which returns a 404 ("resolution not available") if requested image size exceeds the maximum size. (Not sure if 404 is the best solution here)
- Made a change to get_info() which removes the sizeAboveFull value from info.json if max_size_above_full <= 100.
- Added some tests to webapp_t.py

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -16,6 +16,8 @@ In addition to a bunch of directory paths (items that end with `_dp`) which shou
  * `enable_caching`. If `enable_caching=False` no Memory or filesystem caching will happen and `Last-Modified` headers not be sent. This should really only be used for testing/development/debugging.
  * `redirect_canonical_image_request`. If `redirect_canonical_image_request=True` and the request for an image is not the canonical path (e.g. only a width is supplied and the height is calculated by the server), the client will be redirected a `301`.
  * `redirect_id_slash_to_info` If True, `{id}/` and `{id}` will both redirect to the `{id}/info.json`. This is generally OK unless you have ids that end in slashes.
+ * `max_size_above_full` A numerical value which restricts the maximum image size to `max_size_above_full` percent of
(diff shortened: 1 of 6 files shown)


jpstroop:
Wow this is great!

You might be interested to know that the (DRAFT) Image API 2.1 adds some additional options for restricting the size of the image that can be requested. See `maxWidth`, `maxHeight`, and `maxArea` and the statement beneath the table under [Profile Description](http://iiif.io/api/image/2.1/#profile-description).

We (the editors of the IIIF spec) didn't come to add those for security concerns--asking for `full/full` of a 100K pixel image even with the best of intentions can wreak havoc!--but you make a really good point.

Let me think for a bit; we may want to refactor your PR to align it with these new features of the spec...this could be viewed as a separate concern, but I'm not sure that it needs to be. If you have thoughts about that I'd be happy to hear them.

I'm also going to open an issue over at IIIF/iiif.io suggesting that we add some language to the next version of the spec about security concerns related to your point.

jpstroop:
OK. After a little thought (and some time off for the holidays), my instinct is to keep this separate from the `maxArea`, `maxWidth`, and `maxHeight` implementation. In addition to your security concern, there are other reasons why users might want to constrain how much the image can be upsampled, e.g., the quality of the image suffers past a certain point, and a smaller image upsampled 200% could potentially fit within the other `max*` constraints.

jpstroop:
Thanks again!

PR #201, opened 2016-01-08
Bugfix for my last pull request (max_size_above_full)

gvasold:
I noticed that the changes I provided with my last pull request could cause problems when generating info.json. This only occures when option max_size_above_full is set to a value <= 100 and only with cached ImageInfos which were created before 'sizeAboveFull' was added to constants.OPTIONAL_FEATURES. 
The fix is simple: I wrapped the call to `remove('sizeAboveFull')` in a `try ... except ValueError` block.


jpstroop:
Oops, sorry, thought I'd merged this already!

PR #206, opened 2016-04-13
Change SimpleFSResolver to handle multiple directories

epoz:
Using 'src_img_roots' setting in config.

Backwards compatible with single 'src_img_root' setting if no 'src_img_roots' found.


loris/resolver.py
```diff
@@ -60,12 +60,20 @@ class SimpleFSResolver(_AbstractResolver):
...
+            self.cache_roots = self.config['src_img_roots']
+        else:
+            self.cache_roots = [self.config['src_img_root']]
 
-    def is_resolvable(self, ident):
+    def get_fp(self, ident):
         ident = unquote(ident)
-        fp = join(self.cache_root, ident)
-        return exists(fp)
+        for directory in self.cache_roots:
```
jpstroop, line 70: Could you find a way to avoid the loop in the simple case (i.e. one cache root?). Loops are notoriously slow in Python, which means you should consider writing the loop as a list comprehension instead.

Additionally, this should be documented in the config file (and point out that the first match wins, in case of a naming collision), and I'd like to see tests that prove that both cases (single root and multiple) can work.

jpstroop:
This seems like a good idea. See my comment inline.

loris/resolver.py
epoz, line 70: Looking at adding some tests now, noticed that the tests contain tab characters in stead of spaces. Will try to see if it is feasible to clean those up.

epoz, line 70: OK, added some tests and removed the tabs from tests/resolver_t.py.

Don't agree with the comment on changing the loop to a list comprehension, or that it matters in terms of performance in this case. The iteration is through a very small number of directories, and then an os.exists is used.

jpstroop, line 70: :ok: I may refactor later (not a big deal) and I think the fact that we're using `exists()` vs. `os.exists()`, which would add an extra function lookup in every iteration of the loop is probably a good start.

jpstroop:
Could you squash you commits, and I'll get this merged?

demiankatz:
@jpstroop, have you seen GitHub's new "squash and merge" option available through the new drop-down menu on the PR merge button? It showed up a couple of weeks ago, and it might serve you well here (though, of course, if this is about preferred workflow rather than simply saving time in the PR merging process, I don't mean to question your practices!)

jpstroop:
I have. Good thought!

PR #208, opened 2016-04-22
Triplestore resolver

scossu:
Add a resolver that looks for an image URI in a triplestore index given an arbitrary UID. This is useful mostly in a Fedora4 + triplestore index scenario, where a client may only know a unique identifier for an image.

### DIFF
--- a/etc/loris2.conf
+++ b/etc/loris2.conf
@@ -80,6 +80,19 @@ src_img_root = '/usr/local/share/images' # r--
 # default_format
 # head_resolvable = False
 
+## Triplestore index resolver.
+#[resolver]
+#impl = 'loris.resolver.TriplestoreHTTPResolver'
+#source_suffix='/content'
+#cache_root='/usr/local/share/images/loris'
+#sparql_endpoint='<Set this to your triplestore query endpoint>'
+#property_uri='<Set this to the URI of a property indicating a unique identifier>'
+#datatype='http://www.w3.org/2001/XMLSchema#string'
+#qry_template='<Use this property to build a custom query template. Please look carefully at the placeholders in the default template.>'
+#user='<if needed else remove this line>'
+#pw='<if needed else remove this line>'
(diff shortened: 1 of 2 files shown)


jpstroop:
Stefano- this is great, and seems like it would be useful for others. Two questions:
1. How F4 specific is this? Should it be named something like `Fedora4TripleStoreResolver` instead? 
2. Is there anything you can test? I know you're inheriting a fair bit from `SimpleHTTPResolver`, but I wonder it there's anything you can test? Note that we already have [responses](https://github.com/loris-imageserver/loris/blob/development/tests/resolver_t.py#L55) as a dependency which might be useful for mocking out a request/response integration test.

scossu:
> How F4 specific is this? 

In theory, any triplestore indexing any image repository can be used. There is nothing Fedora4-specific in the code. I can foresee this being used mostly by Fedora adopters, but I would not exclude that other institutions exposing a SPARQL endpoint might find it useful, so I would encourage a general-purpose application. 

> Is there anything you can test?

Sure. The config options specific to this class, as well as the HTTP call to the SPARQL endpoint, should be tested. http://collection.britishart.yale.edu/sparql may be a good public endpoint to test against (speaking of non-Fedora implementations and custom query). 

Shall I close this PR or let you merge it and add the tests in a separate one?

jpstroop:
> Shall I close this PR or let you merge it and add the tests in a separate one?

If you're anxious I'm OK w/ merging and you tacking on tests later (I'm sure you'll do it, right? :smile:). Or you could just push another commit to this branch and I can merge later on.

scossu:
Oh, no anxiety at all. I can close this PR and send another one when I complete the tests. I wanted to squash the commits anyway.

jpstroop:
:ok_hand:  :+1:

scherztc:
@scossu Did you ever get any test for this resolver?  Have we confirmed that it works with Fedora4?

PR #214, opened 2016-05-02
Added extra log messages so progress of cleaning can be monitored.

demiankatz:
Not sure if we want this on by default or if it should be controlled by a switch, but I find this useful for troubleshooting, so I figured I would share it.

### DIFF
--- a/bin/loris-cache_clean.sh
+++ b/bin/loris-cache_clean.sh
@@ -10,6 +10,9 @@
 
 LOG="/var/log/loris2/cache_clean.log"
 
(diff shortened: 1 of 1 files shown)


jpstroop:
@demiankatz :arrow_up: I must have merged in the wrong order 😦

demiankatz:
Hopefully not too hard to fix -- I'll sort it out after the conference. Thanks!

demiankatz:
@jpstroop, the conflict is now resolved, so this should merge cleanly again. Thanks!

jpstroop:
:+1:

PR #219, opened 2016-06-07
ident_regex configuration

crhallberg:
This PR adds a configuration to the SimpleHTTPResolver.

By setting `ident_regex` the user can compare idents coming through `is_resolvable` to prevent unnecessary network traffic hitting their source server.

### DIFF
--- a/doc/resolver.md
+++ b/doc/resolver.md
@@ -54,6 +54,7 @@ source_suffix=''
 uri_resolvable=False
 head_resolvable=False #DO set this to true if using Fedora Commons 3.8 or later. Earlier versions have a bug for a head response.
 default_format=None #Set this if your HTTP server doesn't populate content-response. An example value might be "jp2".
+ident_regex=False #Set this to a regular expression matching your identifier pattern to reduce unnecessary network traffic on source server
 user=None
 pw=None
 cache_root='<must be configured>'
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -16,6 +16,7 @@ import constants
 import hashlib
 import glob
 import requests
+import re
 
 logger = getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


demiankatz:
We noticed that our Fedora server was getting constantly bombarded with weird IIIF requests coming from Loris. Adding this setting solved the problem.

jpstroop:
This seems fine, but I'd like to know more about the "weird IIIF requests coming from Loris". Are they really random? Coming from crawlers?

demiankatz:
Without this check, every IIIF URL that goes to Loris gets echoed to Fedora because the code is checking to see if it is resolvable. Thus, the "weird IIIF requests" are a side effect of normal Loris usage.

jpstroop:
OK, just making sure. Thanks!

demiankatz:
Glad to help! For what it's worth, this is the root cause of the problem:

https://github.com/loris-imageserver/loris/blob/development/loris/webapp.py#L329

Perhaps there is a more targeted solution to the problem that doesn't require the new config setting we added... but it didn't seem like a bad idea to add ID validation, and it met our needs, so there you go! In any case, thanks for the super-fast merge!

jpstroop:
Unfortunately, there's not an easier way I've been able to find. The problem is that WSGI, per the spec, escapes URIs before Python has access to them. And since identifiers can include slashes, we need to be able to test whether the string you sent is an identifier that includes slashes, and either redirected to `/info.json` if resolveable, `404` if not resolveable, or `400` if a bad request.  My hope was that is_resolvable could be implemented as a HEAD request, but apparently Fedora 3 doesn't support it.

jpstroop:
The ultimate solution would be to adopt a new framework...and I've been looking into Tornado, which I think could have a lot of advantages not only wrt :arrow_up:, but deployment as well. It would basically amount to a rewrite, but Loris seems ripe for a big refactor anyway...we'll see.

demiankatz:
Thanks for the explanation -- that all makes sense! In any case, hopefully the regex option will save some pain in the meantime.

PR #221, opened 2016-06-09
Add ability to explicitly proxy request through loris.

lfarrell:
@jpstroop We figured out how to proxy requests through loris without changing loris or Leaflet-IIIF. However, I went ahead and made the changes you suggested to allow other folks to proxy by directly overriding the path to their info.json files.


jpstroop:
OK, great, thanks!

PR #226, opened 2016-06-20
Unit tests

jrhoads:
Create isolated unit tests for resolvers.
Relies on dictionaries for config.
Does not rely on the rest of the application config.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 
 *.egg-info/
 
+env/
+.cache/
+
 build/
 
 dist/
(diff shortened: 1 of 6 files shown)


jrhoads:
Can run the unittests with
python -m unittest discover -t . -s tests -p *_ut.py

jpstroop:
Nice! Any reason not to include them in `tests.py`?

Also, could you squash?

jrhoads:
No reason I can think of.
I'll work on those things.
-Joseph

jrhoads:
Rebased and added the unit tests to test.py

jpstroop:
Awesome. Thanks!

PR #235, opened 2016-09-20
remove unused test.sh

bcail:


jpstroop:
Good idea. Travis needed this for a while, can't remember why, but I suppose it just causes confusion now. Thanks!

PR #237, opened 2016-09-23
Dissect uri

bcail:


jpstroop:
Cool, that's _much_ better. Thanks!

Issue #239, opened 2016-10-12
TemplateHTTPResolver does not call its parent class initializer correctly


peichman-umd:
The TemplateHTTPResolver in resolver.py does not call its parent initializer correctly. Instead, it calls `super(SimpleHTTPResolver, self).__init__`, which calls its _grandparent_ initializer (i.e., AbstractResolver). As such, it does not get intialized as a SimpleHTTPResolver, and there is duplicated code in the initializer that could be eliminated.

peichman-umd:
Tagging @rlskoeser and @ghukill as they appear the be the main contributors on the TemplateHTTPResolver.

jpstroop:
Closed by #240

PR #164, opened 2015-06-30
TemplateHttpResolver (based on SimpleHttpResolver)

rlskoeser:
I've extend the SimpleHttpResolver to create a template-based http resolver, based in part on the syntax proposed in #98.  Currently it only supports url patterns, no other configurations are per-template.  If you think that needs to be added before this can be merged, I'm willing to add some logic for that but would like some feedback.  I held off on that in part because I wasn't sure of the best config file format to use; it looks like configobj supports nested settings groups, does that make sense here?  I did write unit tests, and documented the configuration in the class docstring.  Please let me know what additional functionality or documentation you think this needs before it's usable.

I had to refactor SimpleHttpResolver in a couple of places to make it easier to extend, so that I could override just the logic that is different and let the base class do all the real work of resolving and caching, so this is probably also of interest to @scande3.

PR #240, opened 2016-10-12
Fixed subclass relationship in TemplateHTTPResolver. (#239)

peichman-umd:
Eliminated duplicate code by correcting the call to the superclass **init** method in TemplateHTTPResolver.

Fix for #239

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -406,7 +406,10 @@ class TemplateHTTPResolver(SimpleHTTPResolver):
         prior to 3.8).  [Currently must be the same for all templates]
     '''
     def __init__(self, config):
-        super(SimpleHTTPResolver, self).__init__(config)
+        # required for simplehttpresolver
+        # all templates are assumed to be uri resolvable
+        config['uri_resolvable'] = True
+        super(TemplateHTTPResolver, self).__init__(config)
         templates = self.config.get('templates', '')
         # technically it's not an error to have no templates configured,
         # but nothing will resolve; is that useful? or should this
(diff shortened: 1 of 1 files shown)


jpstroop:
@rlskoeser and @ghukill it would be good to know if this was a bug or intentional for some reason I'm missing.

ghukill:
Thanks for the heads up.  My contributions were small, and post-creation of the `TemplateHTTPResolver`,  so while I can't speak to the original designs, looks like it makes sense.  Might explain why we needed to [pull in a particular config parameter](https://github.com/loris-imageserver/loris/commit/076e863ed3c289956fb09d6bb72791dbd61be517#diff-0e4e711cf8f48f97f21ec34c7e8351a9) at one point?

peichman-umd:
@ghukill Yeah, I was able to strip out a bunch of code in the initializer that was copied from the SimpleHTTPResolver.**init** method. We should also be able to remove the line you point to as well.

rlskoeser:
Went back to the original pull request (#164) to look.  The only thing I can think of is that perhaps I was intentionally skipping the irrelevant configurations that are initialized in the SimpleHTTPResolver, and/or avoiding the error handling that doesn't apply to the template config.  (Clearly I should have added a comment to that effect if it _was_ intentional, since it's really not obvious.)

It seems like there might be a more elegant way to handle loading variant configs in subclasses than the way this is implemented, but if it works with the change then it sounds like it's not a problem.  (It looks like setting the `uri_resolvable` flag avoids the error that otherwise would have been raised.)

peichman-umd:
@rlskoeser By variant configs do you mean things like varying not just the URL template but also the user credentials or ssl_verify status based on which template is selected? If so, I also have changes to address that, which I will be submitting. If not, I'd be interested to hear what you are thinking about in terms of variation.

peichman-umd:
To elaborate a bit more on the enhancements for switch request options based on selected template, what I have done is created subsections of the `[resolver]` section named after each template, and placed custom request option settings (e.g., `user`, `pw`) in there. The main change in the code is to make the retrieval of the request options dynamic in the TemplateHTTPResolver. I did this by changing the `_web_request_url` method to return a tuple of `(url, options)` instead of just the URL.

I am waiting to see the resolution of this PR before submitting those changes, since it does depend on the TemplateHTTPResolver calling the SimpleHTTPResolver initializer.

PR #241, opened 2016-10-13
Allow override of request info per template.

peichman-umd:
Changed the return signature of `_web_request_url(ident)` to the tuple of `(url, options)`. This allows the TemplateHTTPResolver to query the prefix-specific subsections of the resolver config for `user`, `pw`, and `ssl_check keys`, and use those values to override the defaults set in the main resolver config.


peichman-umd:
@rlskoeser @ghukill @jpstroop Here is the promised PR for per-template overrides of the request info.

jpstroop:
Looks OK to me, but Princeton isn't using this resolver, so I'd appreciate a review from someone else--anyone above or @scande3?

PR #244, opened 2016-10-18
if a '1,' size parameter would cause the height to be less than 1, se…

bcail:
…t it to 1 instead of letting it go to 0 and cause an error


jpstroop:
Makes sense...[pct requests already do this](https://github.com/bcail/loris/blob/ad762872e95535f7dfe8385adc3630e8fd7f0a33/loris/parameters.py#L316-L324) wonder if there's a way to do it once, but happy to merge this for now. Thanks!

PR #246, opened 2016-10-28
Support reading environment variables into config.

peichman-umd:
Implemented by enabling interpolation in the "template" style (e.g., run_as_user='$USER'). When loading the config, the webapp.py copies the os.environ dictionary into the DEFAULT section of the config object.

Consolidated the two places the ConfigObj is created into one read_config method.


jpstroop:
That's a good idea. Interpolation always makes me concerned about security, but this seems fine. We're starting to do a similar thing, here.

Issue #247, opened 2016-10-31
travis build reports success even with an error


bcail:
See https://travis-ci.org/loris-imageserver/loris/builds/171507934 and download the raw log (https://api.travis-ci.org/jobs/171507935/log.txt?deansi=true). There's a test failure at the bottom.

jpstroop:
Wow, that's a little scary. I just pulled the latest development revisions and see this:

```
FAILED (errors=1)
jstroop@ltopstroop1:~/workspace/loris (development)$ echo $?
0
```

I'll make a PR to see if having it return non-0 changes things.

jpstroop:
Closed by #248

PR #248, opened 2016-10-31
tests.py should not return 0 on fail

jpstroop:
Possible fix for #247

### DIFF
--- a/test.py
+++ b/test.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 # test.py
+from sys import exit
 from tests import img_info_t
 from tests import parameters_t
 from tests import resolver_t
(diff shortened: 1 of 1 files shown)


jpstroop:
Yay, it failed!

jpstroop:
I'm going to merge...may as well advertise the truth.

jrhoads (commented):
Looks good.  We were heading to this solution as well.
Thanks for the quick fix.

Question: Should we keep the verbosity so high for the tests we send to travis?  It seems a shame that we cannot view the error in the travis web view of the log.

jpstroop:
> Should we keep the verbosity so high for the tests we send to travis? It seems a shame that we cannot view the error in the travis web view of the log.

I'm happy to have to volume turned down.

jpstroop:
I think this may be why that `test.sh` script existed, now that I think about it.

Issue #249, opened 2016-10-31
HTTP Resolver fail


jpstroop:
@peichman-umd 

```
Traceback (most recent call last):
  File "/home/jstroop/workspace/loris/tests/resolver_t.py", line 278, in test_template_http_resolver
    self.app.resolver._web_request_url('a:foo.jpg'))
  File "/home/jstroop/workspace/loris/loris/resolver.py", line 461, in _web_request_url
    url = self.templates[prefix]['url'] % ident
TypeError: string indices must be integers, not str
```
This was hidden due to #247. Can you get a fix in?

peichman-umd:
@jpstroop Looking at it now.

jpstroop:
Thanks!

peichman-umd:
Opened PR #250 that should fix this. It was a mixture of bad test data and the TemplateHTTPResolver not returning the expected `(url, options)` tuple in cases where the identifier couldn't be resolved.

bcail:
Fixed by PR #250.

PR #250, opened 2016-10-31
Fix failing tests (#249)

peichman-umd:
- Correct the config format in resolver_t.py
- Explicitly return a `(None, {})` tuple from the TemplateHTTPResolver if the template or identifier can't be found. This avoids running afoul of "NoneType cannot iterate" errors that were occurring in the tests.


peichman-umd:
I was unable to run any of the tests that require jp2 transformations; when it attempts to run `kdu_expand` the test process just hangs. But all of the other tests that I was able to run pass.

PR #251, opened 2016-11-02
Refactor dissect uri

bcail:


jpstroop:
@bcail Have you tried this behind a real server? I imagine it's OK, but so much of the backflipping here (years ago) was only evident when running with Apache, it would be good to double check that it still works.

bcail:
I'll test it behind apache, and also fix the conflicts.

jpstroop:
:+1:

bcail:
Updated the PR. Tried it out in our dev environment, found a problem, fixed it (with a test), and then it worked fine. Haven't done a lot of testing, but initial use in our dev environment is fine now.

jpstroop:
Thanks!

PR #254, opened 2016-11-04
Ensure that values the resolver retrieves from its config will be interpolated.

peichman-umd:
Previously, the resolver_config was a copy of the app config "resolver" section, which was silently converting the configobj.Section to a plain dictionary. THerefore, calls to get values from keys in that section were not being interpolated. This fix just passes the unaltered "resolver" section to the resolver.


jpstroop:
:+1:

PR #256, opened 2016-11-17
Invalid jp2

bcail:


bcail:
looks like the build hung and got killed. @jpstroop could you please restart it?

bcail:
thanks.

PR #259, opened 2016-11-23
libkdu option should default to directory

### LINKED ISSUE #179: confusing documentation around --libkdu option for installation?
We had trouble getting Loris to transform .jp2 files, and after some digging and environment printing during runtime, determined it was a path issue. 

The install option flags for setup.py state:
`--libkdu     Path to libkdu.so [Default: /usr/local/lib/libkdu.so]`

Which suggested to us, that given our installation of Kakadu which has the lib file `libkdu_v77R.so`, we should set that flag as, `/usr/local/lib/libkdu_v77R.so`, as the default contains the .so file explicitly.  This resulted in .jp2 files not processing.  When we removed the file from the option and used simply `/usr/local/lib` it works.

What made this confusing, at least for us, was the `--kdu-expand` option has similar language with the binary file included:
`--kdu-expand      Path to the Kakadu executable [Default: /usr/local/bin/kdu_expand]`,
but **does** actually require the binary to be included (e.g. `/usr/local/bin/kdu_expand`).  

Just thought some clarification in the install flags might help.  Otherwise, so far we're pleased as punch with Loris!  Thanks for making it available.

jpstroop:
Fixes #179 for real.

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -17,7 +17,7 @@ These are the options (they can also be viewed by executing `python setup.py ins
 
 ```
  --kdu-expand      Path to the Kakadu executable [Default: /usr/local/bin/kdu_expand]
- --libkdu          Path to __the directory that contains__ libkdu.so [Default: /usr/local/lib/libkdu.so]
+ --libkdu          Path to __the directory that contains__ libkdu.so [Default: /usr/local/lib]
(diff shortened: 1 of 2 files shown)


jpstroop:
@bcail want to try merging this?

PR #262, opened 2016-11-29
Move code for creating derivative cache directory into ImgCache. Catc…

bcail:
…h the exception if the directory already exists, instead of checking if it exists and then trying to create it.

### DIFF
--- a/loris/img.py
+++ b/loris/img.py
@@ -2,9 +2,10 @@
 #-*-coding:utf-8-*-
 
 from datetime import datetime
+from errno import EEXIST
 from logging import getLogger
 from loris_exception import LorisException
-from os import path,sep,symlink,makedirs,unlink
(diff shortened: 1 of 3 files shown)


loris/img.py
```diff
@@ -282,3 +283,15 @@ def get_request_cache_path(self, image_request):
     def get_canonical_cache_path(self, image_request):
         canonical_fp = image_request.canonical_as_path
         return path.realpath(path.join(self.cache_root, unquote(canonical_fp)))
+
+    def create_dir_and_return_file_path(self, image_request):
+        target_fp = self.get_canonical_cache_path(image_request)
+        target_dp = path.dirname(target_fp)
+        try:
```
jpstroop, line 290: Soo much nicer in python 3: `os.makedirs(path, exist_ok=True)` :-(

PR #263, opened 2016-12-05
Resolver refactor

bcail:


loris/resolver.py
```diff
@@ -309,14 +305,9 @@ def cached_files_for_ident(self, ident):
     def in_cache(self, ident):
         cache_dir = self.cache_dir_path(ident)
         if exists(cache_dir):
```
jpstroop: This can just return `exists(self.cache_dir_path(ident))` rather than the `if/else`, right?

loris/resolver.py
```diff
@@ -381,9 +372,13 @@ def copy_to_cache(self, ident):
 
         logger.info("Copied %s to %s" % (source_url, local_fp))
 
-    def resolve(self, ident):
+    def delete_from_cache(self, ident):
```
jpstroop: I guess I'm not sure about this. Right now the cache API doesn't expose any behaviors directly--the API is `resolve` and `is_resolvable`, and files may be moved around as a side-effect, but that's intended to be hidden. Where would this method be exposed? Would we expect other resolvers to implement it? Should we consider pulling out the source image cache into a separate class? What are your thoughts?

loris/resolver.py
bcail: yup, I'll update that.

loris/resolver.py
bcail: Does it make sense for the SimpleFSResolver to purge items? Seems like maybe not, since Loris isn't putting the files there at all, but maybe it does make sense if it's just a cache that's populated by another process.

I think separating out the source image caching from the resolver would be a good thing. If you think so as well, I could remove this delete_from_cache method from this PR, and I/we could try to work on what it would look like to separate out source image caching.

jpstroop: > Does it make sense for the SimpleFSResolver to purge items?

I found myself asking that as well. [Maybe](https://docs.google.com/document/d/1aRUwT0KNYwg9BcUhGsnPomaMNZCXcaZYutfXHTXO4_o/edit)? I can see a use case for sending `DELETE` to `/identifier`, (and then why not `PUT` or `POST` as well?). As the doc I linked to suggests, I've always kind of wanted to do this.

I think moving the source image cache into a separate dict-like class is probably a good first step, and then we could revisit where we want to expose it. I suspect your instinct to do it in the resolver might be right, but that's layering on a lot of new behavior, and we might want to rename the Resolver class to something like `ImageRepository`.

PR #264, opened 2016-12-14
Provide support for TIFF files with the extension "tiff"

joecorall:
Running loris on CentOS 7 I was receiving a 500 server error that stated "Didn't get a source format, or at least one we recognize" when accessing files with the extension "tiff". This code change fixed this issue.

### DIFF
--- a/doc/resolver.md
+++ b/doc/resolver.md
@@ -11,10 +11,23 @@ impl = 'loris.resolver.SimpleFSResolver'
 src_img_root=/usr/local/share/images
 ```
 
-then `app.resolver.resolve(ident)` will return 
+then `app.resolver.resolve(ident)` will return
 
 ```
 /usr/local/share/images/01/02/0001.jp2
+```
+
+### `ExtensionNormalizingFSResolver`
+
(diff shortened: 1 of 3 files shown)


bcail:
Thanks, @jcorall! I'm thinking that maybe this is something that we could handle in the resolver - we could put a method in _AbstractResolver for grabbing the format from the file extension, and there we could handle tif/tiff, jpg/jpeg, ... with tests. Then this method (something like format_from_file_extension()) could be called by the various format_from_ident methods as needed. Thoughts, @jpstroop?

jpstroop:
Agreed--thanks @jcorall! I do think pushing the [`ExtensionNormalizingFSResolver`](https://github.com/loris-imageserver/loris/blob/development/loris/resolver.py#L107-L128) up into the `AbstractResolver` (as @bcail  suggests) might be cleaner.

joecorall:
Thank you @bcail and @jpstroop ! Implementing ExtensionNormalizingFSResolver resolver fixed this issue, and I agree is much cleaner.

I reverted my previous commit, and pushed another commit documenting ExtensionNormalizingFSResolver a little bit instead.

PR #269, opened 2017-01-03
Rotated PNGs have transparent backgrounds

bboro:
I'm not a Python dev by any stretch, but using my knowledge of coding in other languages, and a small kickstart on Python syntax has helped me to get started on putting the transparent backgrounds on PNGs back again. However, this change might be unwanted due to me not understanding the whole project codebase, I apologise if that's the case.

The original commented out code actually works, and I don't know why it was marked "NOT WORKING". All that was missing was to "save" the actual RGBA/LA type images, since the save function calls would not have run if the image mode ends with "A". I just copy-pasted the save function calls to an "else" statement so they work for RGBA/LA mode images too. The code can probably be further refactored so there's no duplication, but I'll leave that to someone else.

When I tested locally, this change has consistently produced a rotated (77 degrees for example), transparent background PNG image from a JPG source.


bcail:
@bboro Thanks for the PR! Would you be able to add some testing for this code? Ideally it could go in tests/transforms_t.py, but if you'd rather, you could add a more integrated test in WebappIntegration in tests/webapp_t.py (submit a request that rotates the png and make sure you at least get a valid response).

If you don't add the tests, I'll plan to do it and merge the PR.

bboro:
@bcail I've updated the PR. 

I've added a new test image test.png which doesn't have any transparency. Upon rotating it 45 degrees, it should get a lot of transparent parts. In the test, I am actually checking if the image has any transparency in the alpha channel band after the rotation, and if there is at least 1 pixel with an alpha channel value of 0 the test passes. Hope this is good enough.

As an assertion, I've tried the same test but with no rotation (0 degree rotate), and the test fails because the image doesn't have any transparent pixels then. So hopefully it should be functional and sufficient as a regression test.

bcail:
@bboro thanks.

PR #270, opened 2017-01-06
Resolvers - normalize and map file extensions by default for all resolvers

bcail:


loris/resolver.py
```diff
@@ -98,34 +106,9 @@ def resolve(self, ident):
...
-# src_img_root = '/cnfs-ro/iiif/production/medusa-root' # r--
-#   [[extension_map]]
-#   jpeg = 'jpg'
-#   tiff = 'tif'
-# Note that case normalization happens before looking up in the extension_map.
-class ExtensionNormalizingFSResolver(SimpleFSResolver):
-    def __init__(self, config):
-        super(ExtensionNormalizingFSResolver, self).__init__(config)
-        self.extension_map = self.config['extension_map']
+        format_ = _AbstractResolver.format_from_ident(ident)
```
jrhoads: I think you can still use ```self.format_from_ident(ident)```

jrhoads: Also calling using _AbstractResolver here explicitly skips over inheritence.

loris/resolver.py
```diff
@@ -98,34 +106,9 @@ def resolve(self, ident):
...
-# extensions that loris wants to see, e.g.
-#
-# [resolver]
-# impl = 'loris.resolver.ExtensionNormalizingFSResolver'
-# src_img_root = '/cnfs-ro/iiif/production/medusa-root' # r--
-#   [[extension_map]]
-#   jpeg = 'jpg'
-#   tiff = 'tif'
-# Note that case normalization happens before looking up in the extension_map.
-class ExtensionNormalizingFSResolver(SimpleFSResolver):
```
jrhoads, line 119: Any concerns that removing this class might break someone's setup if they are just trying to upgrade?

Might be nice to leave a shell of the class here with a deprecation notice.

bcail: @jpstroop what do you think? You want to do deprecations for 2.2, and then remove stuff in 2.3? Note, though - we did already remove python 2.6 support after the 2.1 release without a deprecation notice (unless we revisit that decision).

Jon, looks like when you release you normally document the changes that users should be aware of - would it be helpful to have a document that we update whenever a change like this goes in?

jpstroop: That feels like a breaking (3.x) change to me. I'm headed out for the day, but, briefly, if this class is going away and another class is providing the same functionality, could we leave an alias in place so that the behavior stays the same?

For release notes, I'm really terrible at them, but I try. Could just start a doc somewhere in the repo. That would probably be easier, for sure.

bcail: ok, I'll put the class/alias back.

There's a further question, though. Current functionality allows users to define any mapping they want (extension to image format). In this change, I moved that mapping into constants. Do users still need to be able to define their own mapping?

jpstroop: I'm :ok: with the constants and ignoring any config, so long as the presence of some config is just ignored and doesn't break anything

bcail:
@jpstroop ok, I updated the PR.

loris/resolver.py
```diff
@@ -98,34 +105,18 @@ def resolve(self, ident):
...
-#
-# [resolver]
-# impl = 'loris.resolver.ExtensionNormalizingFSResolver'
-# src_img_root = '/cnfs-ro/iiif/production/medusa-root' # r--
-#   [[extension_map]]
-#   jpeg = 'jpg'
-#   tiff = 'tif'
-# Note that case normalization happens before looking up in the extension_map.
 class ExtensionNormalizingFSResolver(SimpleFSResolver):
-    def __init__(self, config):
```
jpstroop, line 120: Does this not require a call to `super`?

bcail: No, I don't think so. It just a blank class, so it has to get all its methods from its SimpleFSResolver base class. And the tests seem to work fine.

jpstroop: Huh. Cool!

PR #278, opened 2017-02-02
kdu_expand hung processes

bcail:


jpstroop:
Looks cool! Have you made it hang on purpose?

bcail:
I've run the test suite with an invalid kdu_expand command - that makes it hang, and with this change the process correctly times out and keeps going.

Also, in our production environment, I have seen it terminate a process after it timed out - happened 3 times yesterday.

jpstroop:
OK Great!

Issue #295, opened 2017-05-23
undefined variable in resolver.py


mrdale1958:
Line 238 (in _web_request_url) says
Bad URL request at %s for identifier: %s.' % (source_url, ident)
but source_url is undefined.  I suspect it should just be url

The result is a server error instead of a useful error message

jpstroop:
Good catch! I'm sure you're correct. Want to make a PR?

jpstroop:
Closed with #296

PR #296, opened 2017-05-23
Update resolver.py

mrdale1958:
Fixes undefined variable in resolver.py [from issue 295](https://github.com/loris-imageserver/loris/issues/295)


jpstroop:
Thanks!

Issue #297, opened 2017-05-25
Cache stampede leading to corrupted image


giorgiosironi:
We have cases such as:
```
52.56.127.120 - - [25/May/2017:10:19:00 +0000] "GET /journal-cms:collection/2016-02/2df5cbe1-paleontology-elife-collections.png/0,47,1800,507/2046,576/0/default.png HTTP/1.1" 200 917504 "-" "Amazon CloudFront" 0.028
52.56.127.120 - - [25/May/2017:10:19:01 +0000] "GET /journal-cms:collection/2016-02/2df5cbe1-paleontology-elife-collections.png/0,47,1800,507/2046,576/0/default.png HTTP/1.1" 200 2139617 "-" "Amazon CloudFront" 1.793
```
Corresponding to the logs:
```
[54.87.213.201] out: [pid: 5004|app: 0|req: 3025/49232] 52.56.127.120 () {44 vars in 936 bytes} [Thu May 25 10:19:00 2017] GET /journal-cms:collection/2016-02/2df5cbe1-paleontology-elife-collections.png/0,47,1800,507/2046,576/0/default.png => generated 917504 bytes in 27 msecs (HTTP/1.1 200) 5 headers in 381 bytes (3534 switches on core 0)
[54.87.213.201] out: [pid: 5008|app: 0|req: 4320/49238] 52.56.127.120 () {44 vars in 936 bytes} [Thu May 25 10:18:59 2017] GET /journal-cms:collection/2016-02/2df5cbe1-paleontology-elife-collections.png/0,47,1800,507/2046,576/0/default.png => generated 2139617 bytes in 1792 msecs (HTTP/1.1 200) 5 headers in 382 bytes (7631 switches on core 0)
```
From https://github.com/loris-imageserver/loris/blob/development/loris/webapp.py#L626 and https://github.com/loris-imageserver/loris/blob/development/loris/img.py#L287 it seems that the cached result image is created in place. It is possible that this leads to a race where one process can serve a file from the cache before another process has finished writing to it.

If this is accepted as a bug, I can prepare a patch where the file is created at a random path (I would say https://github.com/loris-imageserver/loris/blob/development/loris/webapp.py#L630) and then move into the cache folder. On Unix this operation would be atomic, as long as the temporary and cache directories are on the same volume.

jpstroop:
[...]

PR #300, opened 2017-06-01
Moving file into cache only after it has been fully written

giorgiosironi:
Solves #297

### DIFF
--- a/loris/webapp.py
+++ b/loris/webapp.py
@@ -15,6 +15,7 @@ import random
 import re
 import string
 from subprocess import CalledProcessError
+from tempfile import NamedTemporaryFile
 from urllib import unquote, quote_plus
 
 #3rd party imports
@@ -620,9 +621,8 @@ possible that there was a problem with the source file
         Returns:
             (str) the fp of the new image
         '''
-        # figure out paths, make dirs
-        n = ''.join(random.choice(string.ascii_lowercase) for x in range(10))
-        temp_fp = '%s.%s' % (path.join(self.tmp_dp, n), image_request.format)
+        temp_file = NamedTemporaryFile(dir=self.tmp_dp, suffix='.%s' % image_request.format, delete=False)
+        temp_fp = temp_file.name
 
         transformer = self.transformers[src_format]
         transformer.transform(src_fp, temp_fp, image_request)
--- a/loris/webapp.py
+++ b/loris/webapp.py
@@ -621,19 +621,18 @@ possible that there was a problem with the source file
             (str) the fp of the new image
         '''
         # figure out paths, make dirs
-        if self.enable_caching:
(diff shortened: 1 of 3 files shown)


jpstroop:
@bcail any thoughts? I think I'm 👍

loris/webapp.py
```diff
@@ -621,19 +621,18 @@ def _make_image(self, image_request, src_fp, src_format):
...
         '''
         # figure out paths, make dirs
-        if self.enable_caching:
-            target_fp = self.img_cache.create_dir_and_return_file_path(image_request)
-        else:
-            # random str
-            n = ''.join(random.choice(string.ascii_lowercase) for x in range(10))
-            target_fp = '%s.%s' % (path.join(self.tmp_dp, n), image_request.format)
+        n = ''.join(random.choice(string.ascii_lowercase) for x in range(10))
+        temp_fp = '%s.%s' % (path.join(self.tmp_dp, n), image_request.format)
```
bcail: Could use tempfile.NamedTemporaryFile here. You can pass in the tmp_dp as the prefix, and the format as the extension.

bcail:
Made one suggestion, but then :+1:

loris/webapp.py
giorgiosironi: Will update on Monday, thanks

giorgiosironi: I tried doing this:
```
-        temp_fp = '%s.%s' % (path.join(self.tmp_dp, n), image_request.format)
+        temp_file = NamedTemporaryFile(dir=self.tmp_dp, suffix='.%s' % image_request.format)
+        temp_fp = temp_file.name
```
but several problems came up:
- the file-like object may go out of scope making `temp_fp` the path to a delete file, because it is automatically deleted
- it is not guaranteed that the path for the temporary file can be reused (`Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms`)
- in the caching use case, we need to atomically rename this temporary file into the cached one, it's not cleary how this plays with the deletion-on-garbage-collection
So I would avoid opening this rabbit hole

bcail: For the deletion issues, you can pass delete=False, and then the file doesn't get deleted automatically - you have to do it yourself.

For the file being re-opened while the temp file is still open, the documentation does say that it works on Unix, which is all we support in Loris:
>  (it can be so used on Unix; it cannot on Windows NT or later)

Seems like the problems you ran into are fix-able, and this change would use the function provided by Python for creating a temp file, instead of rolling our own temp file names.

But, if @jpstroop wants to merge like this, I won't try to block the PR.

giorgiosironi: Thank you for the pointers, added a commit doing this.

bcail:
Thanks! Merging.

PR #303, opened 2017-06-08
Documenting atomic move operations requirement

giorgiosironi:
See https://github.com/loris-imageserver/loris/pull/300 and https://github.com/loris-imageserver/loris/issues/297

This has become a requirement, otherwise the file cannot be atomically moved. I thought I'd document it.

### DIFF
--- a/etc/loris2.conf
+++ b/etc/loris2.conf
@@ -97,9 +97,11 @@ src_img_root = '/usr/local/share/images' # r--
 # url = 'http://<server>/fedora/objects/%s/datastreams/%s/content' # as used with delimiter option below
 
 [img.ImageCache]
+# must be on the same volume as tmp_dp, for atomic move operations
 cache_dp = '/var/cache/loris' # rwx
 
 [img_info.InfoCache]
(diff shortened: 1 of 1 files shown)


bcail:
Thanks, @giorgiosironi. Can you confirm my understanding: the danger is that after the temp file has been fully created, while it's being copied into the cache directory, it'll be read by another process before the copy is complete? That is why the rename/mv operation has to be atomic, so that other processes don't see the file in the cache until it's fully moved over?

giorgiosironi:
Sorry I have missed this comment in my inbox - but yes, the reasoning is that if the rename/mv is not atomic another identical request can be served the incomplete cached result for the duration of the operation. On the same volume, it's just a rename while on different volumes there would be a lot of data to copy.

bcail:
Another note: the os.rename() function can/will actually throw an exception if the tmp_dp directory is not on the same filesystem/device as the caches. It's not just that you may give up the atomic move - it'll crash.

giorgiosironi:
This is correct - adding a comment about that

PR #305, opened 2017-06-14
Propagating original status code

giorgiosironi:
Proposal to return the original response code in case of failure of an HTTP resolver to retrieve the original image. 404 masks the original response code, which should not be a sensitive information and is very useful for debugging (e.g. 403 is a permission problem, 503 a timeout and so on)

I found no specific reason for sticking to 404 in past Github issues.

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -323,7 +323,7 @@ class SimpleHTTPResolver(_AbstractResolver):
                 public_message = 'Source image not found for identifier: %s. Status code returned: %s' % (ident,response.status_code)
                 log_message = 'Source image not found at %s for identifier: %s. Status code returned: %s' % (source_url,ident,response.status_code)
                 logger.warn(log_message)
-                raise ResolverException(404, public_message)
+                raise ResolverException(response.status_code, public_message)
 
             extension = self.cache_file_extension(ident, response)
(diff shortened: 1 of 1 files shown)


jpstroop:
This breaks with the spec: http://iiif.io/api/image/2.1/#error-conditions

What about putting a more helpful message in the body of the response instead?

giorgiosironi:
We could definitely improve the message body (spec says to use plain text), although I don't know whether the resolver's response body should be exposed for security reasons? The resolver's response status code is already there.

To be fair the spec does allow for 400, 401, etc. so these codes may be whitelisted to be reported back?

jpstroop:
That's a good point, re: security. I wonder what other users of the HTTP resolver think?

bcail:
We use a gateway in front of Loris, so users don't see the errors Loris returns anyway. As long as the error is either logged or put in the response, we'll be able to troubleshoot.

However, I would think that anyone sending Loris responses directly to users would want to be careful what information that response exposes. It might be good to handle user-facing responses in Loris in one place (ie. webapp.py), so we have only one place to sanitize any errors for returning to users.

azaroth42:
If the remote web server returns a 500, should Loris really also return 500?  Or if the remote returns an authentication error, then there's potential for confusing clients into going to the authentication API patterns incorrectly.

:-1: from me.

PR #307, opened 2017-06-22
Switch to consistent use of lazy logging

alexwlchan:
This means that the string substitution only occurs if the string is being logged everywhere – this reduces the perf impact of debug logging when it’s not turned on.


jpstroop:
That's great, thanks!

PR #308, opened 2017-06-22
Fix an undefined variable error

alexwlchan:
And adds a test for the function in question, which highlights the original error and prevents future regressions.


bcail:
Thanks!

PR #310, opened 2017-06-23
Improvements to the RotationParameter class

alexwlchan:
I was reading the code for the `RotationParameter`, and spotted a bug: if you pass it a non-float value that includes `.[0-9]` like `.` or `2.7.13` or `.0.`, it throws a ValueError instead of a SyntaxException. This patch changes it to always throw a SyntaxException, along with tests for a few of the bad values I found.

Additionally:

* Some fixes to the docstring and slots, which were quite out-of-date.
* I’ve added a test using [Hypothesis](http://hypothesis.readthedocs.io/en/latest/) *(disclaimer: I’m a maintainer)*, which fuzzes `RotationParameter` to try to find a string that causes it to throw something other than a SyntaxException. This was how I discovered the original bug. Happy to remove this if you don’t want the additional test dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ xb526qv4524_05_0001.jp2
 /python*
 
 todo.md
+
+.hypothesis
--- a/tests/parameters_t.py
+++ b/tests/parameters_t.py
@@ -311,26 +311,21 @@ class TestSizeParameter(_ParameterTest):
 
 class TestRotationParameter(_ParameterTest):
 	def test_exceptions(self):
+		bad_values = [
+			'a',
+			'361',
+			'-1',
+			'!-1',
+			'!361',
+			'-0.1',
+		]
 		try:
-			with self.assertRaises(SyntaxException):
-				rp = RotationParameter('a')
-			with self.assertRaises(SyntaxException):
-				rp = RotationParameter('361')
-			with self.assertRaises(SyntaxException):
-				rp = RotationParameter('-1')
-			with self.assertRaises(SyntaxException):
-				rp = RotationParameter('!-1')
-			with self.assertRaises(SyntaxException):
-				rp = RotationParameter('!361')
-			with self.assertRaises(SyntaxException):
-				rp = RotationParameter('-0.1')
+			for value in bad_values:
+				with self.assertRaises(SyntaxException):
+					RotationParameter(value)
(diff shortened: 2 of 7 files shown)


bcail:
@alexwlchan Thanks for your contributions! As far as adding Hypothesis as a dependency, I'm fine with that but it'd be nice to be able to specify that it's just for tests - use a requirements_test.txt or something. But, I do see that mock and responses are already in requirements.txt. Maybe better organization of the dependencies is a future task. @jpstroop any thoughts?

I'll add a comment inline, but we don't support python 2.6 anymore, so we can remove the duplicated assert statements.

tests/parameters_t.py
```diff
@@ -311,26 +314,32 @@ def test_populate_slots_from_bang_wh(self):
...
+					RotationParameter(value)
 		except TypeError: # Python < 2.7
-			self.assertRaises(SyntaxException, RotationParameter, 'a')
-			self.assertRaises(SyntaxException, RotationParameter, '361')
-			self.assertRaises(SyntaxException, RotationParameter, '-1')
-			self.assertRaises(SyntaxException, RotationParameter, '!-1')
-			self.assertRaises(SyntaxException, RotationParameter, '!361')
-			self.assertRaises(SyntaxException, RotationParameter, '-0.1')
+			for value in bad_values:
+				self.assertRaises(SyntaxException, RotationParameter, value)
```
bcail: This section for Python < 2.7 can be removed. We only support 2.7.

alexwlchan:
Okay, I’ve pulled out the Hypothesis dependency into a different file – left the `mock` and `responses` for now, I’m not entirely sure where else they might be used – and removed the outdated unittest helpers.

.gitignore
```diff
@@ -29,3 +29,5 @@ xb526qv4524_05_0001.jp2
 /python*
 
 todo.md
+
+.hypothesis
```
alexwlchan, line 33: This is a directory used for a bit of bookkeeping and maintenance, but its contents aren’t generally human-readable or something that needs to be checked in.

alexwlchan:
I’m trying to rebase this against the current development branch, but GitHub doesn’t seem to be picking up changes to this PR. 🤔

I’m going to close this and open a follow-up PR.

PR #313, opened 2017-06-26
Improvements to the RotationParameter class

alexwlchan:
This is a second attempt at #310, which GitHub isn’t letting me update with review markups/merge conflicts. 😢

Notes from the description on that PR:

> I was reading the code for the `RotationParameter`, and spotted a bug: if you pass it a non-float value that includes `.[0-9]` like `.` or `2.7.13` or `.0.`, it throws a ValueError instead of a SyntaxException. This patch changes it to always throw a SyntaxException, along with tests for a few of the bad values I found.
>
> Additionally:
>
> * Some fixes to the docstring and slots, which were quite out-of-date.
> * I’ve added a test using [Hypothesis](http://hypothesis.readthedocs.io/en/latest/) *(disclaimer: I’m a maintainer)*, which fuzzes `RotationParameter` to try to find a string that causes it to throw something other than a SyntaxException. This was how I discovered the original bug. Happy to remove this if you don’t want the additional test dependency.

### DIFF
--- a/loris/parameters.py
+++ b/loris/parameters.py
@@ -393,6 +393,7 @@ class SizeParameter(object):
     def __str__(self):
         return self.uri_value
 
+
 class RotationParameter(object):
     '''Internal representation of the rotation slice of an IIIF image URI.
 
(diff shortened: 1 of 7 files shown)


loris/parameters.py
```diff
@@ -393,44 +393,56 @@ def __mode_from_size_segment(size_segment):
...
+
     Slots:
-        uri_value (str)
         canonical_uri_value (str)
-        mirror (str)
+        mirror (bool)
         rotation (str)
     '''
-    ROTATION_REGEX = re.compile('^!?[\d.]+$')
+    ROTATION_REGEX = re.compile(r'^(?P<mirror>!?)(?P<rotation>[\d.])+$')
```
alexwlchan: oops, the `rotation` group only captures a single digit

bcail:
Thanks.

PR #314, opened 2017-06-27
Reduce a bit of code repetition around directory creation

alexwlchan:


bcail:
I see you restored the _create_cache_dir method. It looks to me like it could just be removed, along with its test. Is there a reason to still keep it around?

alexwlchan:
> Is there a reason to still keep it around?

Yeah, now I look again, I can’t think of one. (I hadn’t noticed it was unused after my patch.) I’ll remove it.

PR #316, opened 2017-07-06
Add Authentication support to Loris

azaroth42:
Makes some significant changes, notably passing around the info object rather than just variables.


authnServices.py
```diff
@@ -0,0 +1,283 @@
+from bottle import request, response, abort, redirect
+from bottle import auth_basic, parse_auth, Bottle, run
+from bottle import debug as set_debug
+import json
+import sys
+
+# XXX These probably aren't Python 3.x safe?
```
alexwlchan: I think the `netaddr` import is fine (that’s a third-party lib), `urllib` less so. What most people do is either use `six.moves` or import the specific functions you use. Something like:

```python
try:
    from urllib import urlencode
    from urllib2 import Request, urlopen
except ImportError:  # Python 3
    from urllib.parse import urlencode
    from urllib.request import Request, urlopen
```

azaroth42: Thanks @alexwlchan! Changes made.

azaroth42:
Yes, the tests fail. LMK about the scope of changes, and if they're okay, I'll fix up the tests.

bcail:
Thanks - I think it'll be great to support the IIIF authentication API.

Some questions:
1. do we have to add bottle as a dependency, or could we just use werkzeug instead?
2. could you add good tests for any new code?
3. could the S3CachingResolver (or anything else) be split out into a different PR?

azaroth42:
1. Yes, we can use werkzeug (though PRs very welcome for the probably few changes needed). The authnServices.py is just to show that Loris doesn't need to be responsible for authentication at all, just the redirection process. We don't even run them on the same server :)  It's intentionally stand alone so that institution-specific authN/Z business logic doesn't affect the image serving code.
2. Yes, will do once there's agreement that the changes are justified and the right direction. @jpstroop was hoping to just put the authN into resolvers without a new class, but I couldn't get the workflow right... that could easily be my misunderstanding of the patterns though, and a quick refactor might be able to merge them.
3. Good call. Will split the S3 stuff out, it's completely incidental to the auth changes.  It was in there because I had to develop the IP based authentication pattern against our live system (bleh! 😖 ).

azaroth42:
Tests are now green (they found one real bug, where I'd forgotten to move the catch for ResolverException) and will work on tests for authorizer.

azaroth42:
@alexwlchan @jpstroop @bcail Ready to go?

bcail:
@azaroth42 thanks for your work on this. I'm planning to look more at it next week.

bcail:
@azaroth42 I see that you're getting the <source_fp>.rules.json file in the SimpleFSResolver. Loris has the SimpleHTTPResolver & TemplateHTTPResolver as well, where source image files are grabbed from an external service - how is it supposed to retrieve the rules.json file in those setups?

bcail:
@azaroth42 ping. Any thoughts on generalizing the rules-based authentication for all the Loris resolvers? Or would it only be supported for the FS resolver?

azaroth42:
Sorry, back at work tomorrow and will look.

azaroth42:
Green again with HTTP resolver implementation for rules file.

bcail:
@azaroth42 Thanks, I'll take a look tomorrow.

authnServices.py
```diff
@@ -0,0 +1,314 @@
...
+        self.AUTH_URL_LOGOUT = "logout"
+
+        # login:
+        # self.handler = BasicAuthHandler(...)
+        # self.handler = OAuthHandler(...)
+        # self.handler = IPRangeHandler(...)
+        self.handler = BasicAuthHandler(self, cookie_secret, token_secret, salt, name,
+            cookie_domain, cookie_path, is_https)
+
+    def send(self, data, status=200, ct="text/plain"):
```
alexwlchan, line 53: nit: Any reason for brevity? Why not call this `content_type`?

authnServices.py
```diff
@@ -0,0 +1,314 @@
...
+        response.headers['Access-Control-Allow-Origin'] = '*'
+        response.headers['Access-Control-Allow-Methods'] = "GET,OPTIONS"
+        response.headers['Access-Control-Allow-Headers'] = 'Origin, Accept, Authorization'
+        response["content_type"] = ct
+        response.status = status
+        return data
+
+    def not_implemented():
```
alexwlchan: minor: Is this missing a `self` parameter?

bcail:
@azaroth42 I'm going to give some more feedback. If you want to give me access to your fork, I can help work on it.

requirements.txt
```diff
@@ -2,3 +2,6 @@ werkzeug >= 0.11.4
 pillow >= 2.4.0
 configobj >= 4.7.2,<=5.0.0
 requests >= 2.12.0
+mock == 1.0.1
```
bcail: this should go in requirements_test.txt

loris/resolver.py
```diff
@@ -340,14 +368,36 @@ def copy_to_cache(self, ident):
                 rename(tmp_file.name, local_fp)
                 logger.info("Copied %s to %s", source_url, local_fp)
 
+        # Check for rules file associated with image file
+        # These files are < 2k in size, so fetch in one go.
+        # Assumes that the rules will be next to the image
+        # cache_dir is image specific, so this is easy
+
+        bits = split(source_url)
+        fn = bits[1].rsplit('.')[0] + "." + self.auth_rules_ext
```
bcail, line 377: Seems like source_url would be something like http://mysite.com/images/foo.jpg, so split() would just return ['http://mysite.com/images/foo.jpg'], and bits[1] would give an IndexError. We could factor out a _rules_url(source_url) method and test it with various inputs, and make sure it's returning what we want.

loris/resolver.py
```diff
@@ -61,13 +65,40 @@ def resolve(self, ident):
             ident (str):
                 The identifier for the image.
         Returns:
-            (str, str): (fp, format)
+            ImageInfo: Partially constructed ImageInfo object
```
bcail, line 68: Wouldn't it be better to construct the ImageInfo object in one step, instead of having a partially constructed object? We could pass the transformers from webapp.py into resolve(), which would let us do the whole thing in one step. Or, we could have resolve() not return the ImgInfo object, but just return the additional path to the rules file (if it exists), and then the ImgInfo object could be constructed outside of the resolvers. Thoughts?

azaroth42: The partially constructed object is still useful for some operations, notably testing whether the request is authorized which needs a lot of pointers but not the actual pixel information.  The other case is the info cache, but there it's constructed via from_json(). There are a few tests that occur between constructing the info object and loading pixels if needed, which are different in get_info and get_img.

As to the suggestions, yes, they could work. My opinions:

* Returning an array of variables rather than an object is (IMO) an anti-pattern in any object oriented paradigm. We'll always run into situations where we need just one more piece of information that could easily (and probably already is) on the info object. I can see the point of @jpstroop's tuple, but once you start needing more, my experience is that it's a very slippery slope into needing everything.
* I worry about overloading the resolve() to do lots of testing (per get_info and get_img) that need application level configuration information. That also seems like a slope towards a single function that does everything (which I did in my first implementation of the API, and was terrible).

`from_image_file()` returns self, so there's a reasonably convenient pattern if you have all the information and just need the whole thing:

`info = ImageInfo(ident, fp).from_image_file()`

loris/resolver.py
azaroth42: Split here is `os.path.split()`, meaning:

```
>>> from os.path import split
>>> split("http://example.org/path/to/test.jp2")
('http://example.org/path/to', 'test.jp2')
```

requirements.txt
azaroth42: Probably a bad conflict merge -- I don't use mock. Will delete it.

loris/resolver.py
azaroth42: Hmm, I take that back. _get_info just has a little test between resolve() and from_image_file() that we should be able to move to the constructor. Will try it...

azaroth42: Okay, I'm at a loss for why the final two tests aren't working. I believe it's something to do with the testing framework, rather than the code.  Have added you (@bcail) to my repo if you could take a look at the errors?

authnServices.py
```diff
@@ -0,0 +1,317 @@
...
+        elif domain[-1].isdigit():
+            # 10.0.0.1
+            secret = origin
+        elif len(domain) >= 2:
+            # foo.gtld
+            secret = ".".join(domain[-2:])
+        else:
+            # localhost or * ... hopefully not *!
+            secret = domain[0]
+        return secret
```
alexwlchan, line 134: This seems quite complicated and fragile – is there a reason you’re not just using `urlparse` from the standard library? At least for the test cases defined below, this code does the same thing:

```python
@staticmethod
def basic_origin(origin):
    u = urlparse(origin).hostname
    return u[len('www.'):] if u.startswith('www.') else u
```

alexwlchan (commented):
A few very minor comments from a quick skim.

authnServices.py
azaroth42: No reason :) Can change it. Used to kwargs, not keyword_arguments.

authnServices.py
azaroth42: This is inherited from previous codebases, will test both ways

authnServices.py
azaroth42: Will add more tests. The intent is to shorten x.y.z.foo.com to foo.com, but x.y.z.foo.co.uk to foo.co.uk.  There's a much more sophisticated algorithm that browsers do with the public suffix list, but this is fast and good enough. 
See:  https://en.wikipedia.org/wiki/Public_Suffix_List 

Will add more tests to demonstrate the requirements.

loris/resolver.py
bcail: fixed the issues. The tiff content that was being passed for the resolver tests wasn't valid for Pillow's code.

azaroth42: Ahha, thank you!  Back to green.

After a bit of discussion with our new Enterprise Software Architect, we decided to replace the current, custom cryptography code with a higher level spec -- JSON Web Tokens.  I haven't done this yet, but should get it finished today.

authnServices.py
azaroth42: Updated to use urlparse() with some tweaks, plus the tests for the PSL.

azaroth42:
We have this version (as of commit dedcdfd) running in production now, and works as expected with the digirati auth viewer.

bcail:
@azaroth42 does this PR fully implement the IIIF Auth 1.0 API?

I have more suggestions for changes, but if there are no objections, I may go ahead and merge it and work can continue with other PRs.

azaroth42:
Thanks @bcail! I believe that it does implement everything in the API. It works with the reference client implementation, for all of the patterns. Lacking another validation process for Auth, that seems as good a demonstration of completeness as currently possible.

bcail:
Merged. Thanks for your work.

azaroth42:
Thanks for the merge! We should start creating more granular issues for improvements :)

PR #319, opened 2017-07-08
Slightly better exception handling

alexwlchan:
* Make sure KeyError includes the value that caused the error
* A `try … catch` block that just re-raises the error can be dropped with no effect


bcail:
Thanks!

PR #322, opened 2017-07-10
Add a test and bugfix for InfoCache

alexwlchan:
Linting spotted that `_get_color_profile_fp` isn’t a defined method on InfoCache. I wrote a test for the `__delitem__` method that would expose the bug.

Along the way, I uncovered a second bug in that method, and made a small simplification in `webapp_t.py`.

### DIFF
--- a/tests/img_info_t.py
+++ b/tests/img_info_t.py
@@ -10,6 +10,9 @@ from werkzeug.datastructures import Headers
 import json
 import loris_t
 
+from werkzeug.test import EnvironBuilder
+from werkzeug.wrappers import Request
+
 
 """
 Info unit and function tests. To run this test on its own, do:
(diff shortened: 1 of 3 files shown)


tests/img_info_t.py
```diff
@@ -344,6 +347,27 @@ def test_info_goes_to_http_fs_cache(self):
...
+    def test_can_delete_items_from_infocache(self):
+        '''
+        Test for InfoCache.__delitem__.
+        '''
+        cache = img_info.InfoCache(root=self.SRC_IMAGE_CACHE)
+
+        path = self.test_jp2_color_fp
+        builder = EnvironBuilder(path=path)
+        env = builder.get_environ()
+        req = Request(env)
```
bcail: could you use webapp_t._get_werkzeug_request() here?

alexwlchan: Done.

tests/img_info_t.py
```diff
@@ -10,6 +10,11 @@
 import json
 import loris_t
 
+from werkzeug.test import EnvironBuilder
+from werkzeug.wrappers import Request
```
bcail: looks like these 2 imports can be removed now, right? Then I'll merge.

alexwlchan: Done.

bcail:
Thanks!

Issue #324, opened 2017-07-12
Python 3 compatibility


rogerhoward:
I'm interested in seeing loris support Python 3 (3.5+) and am willing to tackle this, having quite a bit of experience over the past few years migrating old projects forward. Apologies if this has already been discussed and ruled out - I searched old issues without finding anything.

Before I embark on this, I'd like some input, or certainly get any *over my dead body* comments out of the way first. There are several approaches this could take, in no particular order:

1. Update to Python 3-only - this is easiest on the person doing the porting, but probably a non-starter as it would force all current users to update their environments.
2. Update to Py2/3-compatible code, using `six` where useful to abstract some of the more annoying differences. This is the route most projects I'm familiar with have taken - eg., Django. This would allow most current deployments to update to a new Python 3-compatible version without having to update to Python 3.
3. Update to Py2/3-compatible code without using `six` - avoids a dependency, but I don't think this gains much, ends up reimplementing six but worse.

Reviewing the current project, the majority of changes are pretty straightforward: dealing with string vs. unicode literals, print statements, some module renaming (esp. urllib), and fixing relative imports. 

Because this change will (likely) touch every .py file in the project, it'll require some coordination to merge.

Thoughts? Objections?

jpstroop:
FWIW I'm +💯 for this, probably supporting 2.7 and >3.5 for now. I think
the hardest part might be the jp2 info extraction, but I've already done
it! I started a major refactor last summer in a separate repo that kind of
fizzled (though it was basically done):

https://github.com/jpstroop/loris-redux/blob/master/loris/info/jp2_extractor.py

It's also much nicer code with better tests, if I say so myself.

-Js

Sent via mobile. Please excuse typos, brevity, etc.

rogerhoward:
Excellent - was planning to target 2.7/3.5

I think it'd be best to keep the refactor and this compatibility push separate - I'm a little afraid of the refactor without better test coverage (re: #325 and @alexwlchan comments). The 2/3 compatibility is immediately useful, low risk, and a relatively easy job (unsexy as it may be).

This is also a good chance for me to get more familiar with the code/architecture, while I help out on the outskirts.

jpstroop:
Yeah, I wouldn't use all of that code, just the jp2 parser: https://github.com/jpstroop/loris-redux/blob/master/loris/info/jp2_extractor.py#L103-L210. There wasn't a straight 2 to 3 upgrade there because of the something to do the `struct` module, if I recall.

[...]

PR #325, opened 2017-07-12
Start collecting coverage data in CI

alexwlchan:
I am 💯 on #324, and have a few other ideas for more substantial changes, but for major changes it would be good to have an idea of the state of testing (and more specifically, what _isn’t_ tested).

This patch adds coverage information to Travis outputs. This can be used to find parts of the codebase that still need testing.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,7 @@
+[run]
+branch = True
+include =
+    loris/*.py
+
+[report]
+show_missing = True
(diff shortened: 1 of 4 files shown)


rogerhoward:
Totally agree - I'd be reluctant to embark on major refactoring without assessing test coverage for impacted areas first. #324 isn't zero risk, but it's relatively straightforward - no architecture or API changes, for instance.

bcail:
Thanks!

PR #330, opened 2017-07-14
Switch to using py.test as the test runner

alexwlchan:
This patch swaps out the test runner to be py.test, not unittest. Using py.test has a number of nice features:

* Cleaner, more Pythonic asserts
* Parameterized tests
* Less testing boilerplate

For now, I’ve just flipped the test runner – converting the test code at the same time would be a much more unwieldy patch.

While I was poking test infrastructure, a few other changes:

* Don’t install the mock and responses dependencies by default – responses only when we‘re running tests, mock not at all (turns out, it’s never used in the current test suite)
* Installing from `requirements_test.txt` now gets you everything in `requirements.txt` as well, so there’s one less command to run
* Developer docs have been updated with new instructions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ install:
  - pip install -r requirements_test.txt
 
 script:
-  - "coverage run -m py.test tests/*.py"
+  - "coverage run --module py.test --verbose tests/*.py"
   - "coverage report"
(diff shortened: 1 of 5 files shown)


bcail:
Thanks!

PR #344, opened 2017-08-25
Added extra info about AllowEncodedSlashes

ColeDCrawford:
Ubuntu doesn't inherit this for VirutalHosts if set in a global context, so copy and pasting the suggested config block did not work for me. Adding extra info for similar environments.

### DIFF
--- a/doc/apache.md
+++ b/doc/apache.md
@@ -45,7 +45,7 @@ Explanation:
 
  (Loris is setting the `Last-Modified` header based on the file system metadata.)
 
- * `AllowEncodedSlashes On` lets `%2F` though in requests (they're allowed but must be escaped in the identifier portion of the URI).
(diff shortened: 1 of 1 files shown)


bcail:
Thanks, Cole. What version of apache are you on? Seems like maybe this is fixed in more recent versions of apache? If so, could you put in the version you're on where it didn't work?

ColeDCrawford:
I'm on the last release of Xenial (16.04.3). I know the Apache bug says that it's closed, but I think it was really just a documentation update and didn't change the way AllowEncodedSlashes works. I haven't tested on any 17 releases though.

bcail:
OK, so it's still an issue on apache 2.4. I'll go ahead and merge. Thanks for the PR.

PR #350, opened 2017-09-06
Update Loris to allow converting embedded colour profiles to sRGB on non-JP2 image formats

alexwlchan:
This patch is for addressing https://github.com/wellcometrust/platform/issues/734. We noticed that Loris was stripping the Kodak ProPhoto colour profile from transformed JPEGs, so they looked wrong.

Loris already has support for converting the embedded colour profiles on JP2 images to sRGB. The headline feature of this patch is extending that support to other formats.

This involved making the profile conversion more generic, and I made a bunch of other small changes along the way – mostly aimed at making Loris give more useful errors if you’ve configured this feature incorrectly, so you fall over at startup rather than with an AttributeError during image conversion.

Still to do:

- [x] Add some tests of the embedding process
- [x] (Maybe) Pull the imports into the mixin, so they only run when really needed. Reduces the risk of irrelevant ImportErrors, and would slightly improve performance when this feature wasn’t enabled

Those aside, I think this PR is mostly ready for review.


loris/transforms.py
```diff
@@ -131,42 +184,34 @@ def _derive_with_pil(self, im, target_fp, image_request, rotate=True, crop=True)
...
     '''
     def __init__(self, config):
-        self.map_profile_to_srgb = bool(config['map_profile_to_srgb'])
         self.mkfifo = config['mkfifo']
         self.tmp_dp = config['tmp_dp']
 
-        if self.map_profile_to_srgb and \
-            ('PIL.ImageCms' not in sys.modules and 'ImageCms' not in sys.modules):
-            logger.warn('Could not import profileToProfile from ImageCms.')
-            logger.warn('Images will not have their embedded color profiles mapped to sSRGB.')
```
alexwlchan: I think this warning was unhittable – because `profileToProfile` is imported at the top of this file, if the libraries aren’t installed, we’d never reach this point in the file.

bcail:
Thanks, @alexwlchan. Let me know when you're ready for me to review. If you can keep the PR from getting too big with unrelated changes, that'd be great.

alexwlchan:
Okay @bcail, I think this is ready for review. I’ve tidied the commit history and spun out what I can into separate PRs. Thanks for the speedy review on those!

Of particular note:

- I added tests in the style of #351 for JPEG-via-Pillow and JP2-via-OpenJPEG. The latter took a bit of extra work to get right, but I wanted to be sure I hadn’t broken anything.
- I switched from `StringIO` to `BytesIO` because my tests were getting a TypeError otherwise. I’m not entirely sure why that’s the case.
- The original PR was quite mixin and subclass heavy; I’ve tried to pull back from that because I don’t think we need it here.

PR #354, opened 2017-09-12
Bugfixes from the color profile conversion work

alexwlchan:
This patch corrects two issues we discovered after deploying the color profile conversion in #350. Specifically:

* If the image is rotated, make sure we still convert the profile directly.
* If conversion fails, just log a warning and return an image with a potentially incorrect profile, rather than failing entirely. This was caused by issues we had when trying to view CMYK images through Loris.

@bcail This patch is (now) ready for review.


alexwlchan:
Swapping this to WIP while I sort out the test failures. Plus I might need a second patch around failed transforms.

PR #355, opened 2017-09-13
Unrecognised idents in TemplateHTTPTransformer should be 404, not 500

alexwlchan:
This resolves an issue originally raised as https://github.com/wellcometrust/platform/issues/786

TemplateHTTPResolver uses `_web_request_url` to get the source URL for an ident; if it doesn't find something that matches one of the templates, it returns None. It then attempts to download the source image as `requests.get(None)`, which throws a MissingSchema exception. This bubbles up and manifests as a 500 error to the user.

The correct response is a 404 – Loris is unable to find the ident in question.

This patch throws a 404 in the case where the resolver can't pick a source URL, and adds a test for this path. It also skips creating the cache directory until we have a plausible source URL, so errors return slightly faster.

This patch is ready for review.

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -339,8 +339,6 @@ class SimpleHTTPResolver(_AbstractResolver):
 
     def copy_to_cache(self, ident):
         ident = unquote(ident)
-        cache_dir = self.cache_dir_path(ident)
-        mkdir_p(cache_dir)
 
         #get source image and write to temporary file
         (source_url, options) = self._web_request_url(ident)
@@ -353,6 +351,9 @@ class SimpleHTTPResolver(_AbstractResolver):
             )
             logger.warn('Unable to determine source URL for ident %r', ident)
             raise ResolverException(404, public_message)
+
+        cache_dir = self.cache_dir_path(ident)
+        mkdir_p(cache_dir)
 
         with closing(requests.get(source_url, stream=True, **options)) as response:
             if not response.ok:
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -404,36 +404,59 @@ class SimpleHTTPResolver(_AbstractResolver):
 
 
 class TemplateHTTPResolver(SimpleHTTPResolver):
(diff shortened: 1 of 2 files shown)


bcail:
@alexwlchan In SimpleHTTPResolver, _web_request_url verifies that it has a valid URL - if it doesn't, it raises the 404 Resolver exception. Basically, it does the check there instead of in copy_to_cache. But TemplateHTTPResolver doesn't raise exceptions on invalid urls in _web_request_urls - it returns None values. Would it work to make TemplateHTTPResolver more like SimpleHTTPResolver and do the check for a valid URL in _web_request_url?

alexwlchan:
> Would it work to make TemplateHTTPResolver more like SimpleHTTPResolver and do the check for a valid URL in _web_request_url?

Yes, I think that would work, and arguably makes more sense – if `_web_request_url` returns `None`, there isn't anything a caller can useful do except throw a 404. I’ll modify the patch.

alexwlchan:
Okay, now `_web_request_url` raises a ResolverException if it can’t find a URL. I think this is ready for another review.

I’ve also written a bunch of new tests, because I didn’t want to dive into this refactor without a safety net. :-)

loris/resolver.py
```diff
@@ -401,36 +404,59 @@ def resolve(self, app, ident, base_uri):
...
+    The configuration may also include the following settings, as used
+    by SimpleHTTPResolver:
+
+    *   ``default_format``, the format of images (will use the Content-Type
+        of the response if unspecified)
+    *   ``head_resolvable`` with value True, whether to make HEAD requests
+        to validate object existence (don't set if using Fedora Commons
+        prior to 3.8.)  [Currently must be the same for all templates.]
+
+    """
```
alexwlchan, line 459: Along with some examples, big change here is that the `delimiter` option is now documented – previously it was used in code but never mentioned in docs.

bcail:
Thanks for the fix, and the tests.

PR #364, opened 2017-09-30
Improve line coverage of transformer.py

alexwlchan:
Lots of low-hanging fruit to pick off here, but something funky is happening with the WEBP test.


alexwlchan:
The WebP test failure looks like an upstream bug, not a Loris issue, so I've changed the way our tests check images to workaround it.

That aside, I think this is ready for review. :-)

Issue #356, opened 2017-09-14
ImageInfo architecture


bcail:
Currently, the ImageInfo class serves 2 purposes:
1. It's the IIIF info that's returned to the user. https://github.com/loris-imageserver/loris/blob/development/loris/webapp.py#L511
2. It's information about an image that Loris uses internally. https://github.com/loris-imageserver/loris/blob/development/loris/webapp.py#L637 (src_img_fp, src_format).

Should these purposes be split out into different classes, or separated somehow? Am I missing something?

Thoughts?

bcail:
I'm currently seeing errors in our Loris instance where the src_format is '', and that's causing a KeyError here: https://github.com/loris-imageserver/loris/blob/development/loris/webapp.py#L688.

bcail:
Looks like ImgInfo.to_dict() gives you the IIIF info. ImgInfo.to_json() dumps that to_dict data as json if cache param is False, and adds in extra Loris-specific specific information if cache param is True. That's how ImgInfo handles its dual purposes.

I think I'll try refactoring it some, to make things clearer and hopefully fix the issue I'm seeing.

alexwlchan:
> Should these purposes be split out into different classes, or separated somehow?

If I had the time, I’d split it into two classes – one for internal logic, one for display logic. The latter moves into `webapp.py`, the former remains in `img_info.py`. No code is shared between them, just data.

The internal class is what’s left of the old class, and splitting out the display logic makes that class very easy to test – because it only ever has to pass around data in-memory, rather than also try to read stuff from files on disk. Data-only functions are usually easier to test and fuzz.

jpstroop:
Have a look at this: 

https://github.com/jpstroop/loris-redux/blob/master/loris/info/jp2_extractor.py

bcail:
Thanks for your comments, @alexwlchan and @jpstroop.

I think the current issue I have is with the cache. src_format is now in the ImageInfo object, which means it could come from the ImageInfo cache. However, I have old objects that were already cached without the src_format information, and I think that's why I'm getting some KeyErrors on src_format==''.

One solution is to require everyone to purge their ImageInfo cache when they upgrade. Or we could add something to the code to try to force-override objects in the cache with the new information, although that would complicate things in the code. Or we could look at separating src_format from the ImageInfo that would come back from the cache.

Thoughts on any of this? Is it an option to have everyone purge their ImageInfo cache when they upgrade, or would another solution be better?
[...]

PR #368, opened 2017-10-10
Check ImgInfo data from cache. Update ImgInfo method names.

bcail:


bcail:
@alexwlchan could you please take a look at this when you get a chance?

loris/img_info.py
```diff
@@ -315,8 +314,8 @@ def sizes_for_scales(self, scales):
     def scale_dim(dim_len, scale):
         return int(ceil(dim_len * 1.0/scale))
 
-    def to_dict(self):
-        logger.debug('self.ident in to_dict: %s', self.ident)
+    def _get_iiif_info(self):
+        #returns only IIIF info (not Loris-specific info like src_format)
```
alexwlchan: This should be a docstring, not a comment.

loris/img_info.py
```diff
@@ -337,21 +337,18 @@ def to_dict(self):
...
-        if cache:
-            # Add in internal properties for caching
-            d['_src_img_fp'] = self.src_img_fp
-            d['_src_format'] = self.src_format
-            d['_auth_rules'] = self.auth_rules
+    def to_iiif_json(self):
+        d = self._get_iiif_info()
+        return json.dumps(d)
 
+    def to_full_info_json(self):
```
alexwlchan, line 344: Might be worth a brief docstring explaining this method is creating the data that gets cached in the InfoCache.

loris/webapp.py
```diff
@@ -517,7 +517,10 @@ def _get_info(self,ident,request,base_uri):
         else:
             in_cache = False
 
-        if in_cache:
+        #Checking for src_format in ImageInfo signals that it's not old cache data:
+        #   src_format should be in the cache now.
+        #   TODO: remove in Loris 4.0.
+        if in_cache and self.info_cache[request][0].src_format:
```
alexwlchan: A few questions:

* The old cache data was `info.to_json(cache=True)`, which would presumably include `src_format`? So is this addressing an earlier time where it wasn't included?

* As Loris is 2.1.0 now, next major release is 3.0 – so why wait until 4.0 to make breaking cache changes?

* To be clear: what we'd be removing in 4.0 is the second half of this check? Maybe that should be explicit in the comment?

Meta-comment: I didn't immediately recognise the format of `self.info_cache[request]` – I wonder whether we could store namedtuples in InfoCache for greater clarity here?

alexwlchan (commented):
Disambiguating `to_iiif_json()` and `to_full_info_json()` is a nice change. A few questions inline, but nothing too major.

loris/webapp.py
bcail: @alexwlchan thanks for your review. I pushed another commit. As far as whether to remove the cache check in 3.0 or 4.0, @giorgiosironi had a comment about it in https://github.com/loris-imageserver/loris/issues/356#issuecomment-330505838. @giorgiosironi what are your thoughts about removing the cache check in a 3.0 or a 4.0 release?

I'm OK either way. Regardless, we'll have to let users know that they'll have to make sure their cache doesn't have any of the old content once we remove the check.

alexwlchan: Ah, I’d missed the idea of making 3.0 a Python 3-compatible release, and deferring the cache changes until later. I don’t mind too much when we change it, I just wanted to check if I’d forgotten another conversation, and I had.

Important thing is that we have a note to come back and change this later, which we do! :-)

alexwlchan (approved):
Okay, LGTM. And… ✨ merging ✨ .

PR #370, opened 2017-10-13
More human-readable errors around the extraInfo parameter on ImageInfo

### LINKED ISSUE #369: How is the extraInfo parameter in ImageInfo meant to work?
There’s a snippet of code in the constructor for ImageInfo as follows, which appears to pass the value of an `extraInfo` key through directly:

```python
for (k, v) in extra.get('extraInfo', {}).items():
    setattr(self, k, v)
```

I’m trying to add a test for this code, because it’s not currently exercised by the tests. There’s a comment in #298 which describes the intended use case:

> Each key, value in `extraInfo` would be copied into the info.json verbatim.

I understood this to mean arbitrary key-value pairs, but trying that gets an AttributeError. For example:

```pycon
>>> from loris.img_info import ImageInfo
>>> ImageInfo(extra={'extraInfo': {'foo': 'bar', 'bar': 'baz'}})
  File "<stdin>", line 1, in <module>
  File "loris/img_info.py", line 82, in __init__
    setattr(self, k, v)
AttributeError: 'ImageInfo' object has no attribute 'foo'
```

Looking at the [IIIF Image spec](http://iiif.io/api/image/2.1/#table-of-contents), it doesn’t look like there’s support for arbitrary parameters in `info.json`, so is this just intended to be a passthrough for the `attribution`, `license`, `logo` and `service` fields?

/cc @azaroth42 as the original author of this code

alexwlchan:
Resolves #369.

This catches the AttributeError, and rethrows it as an ImageInfoException with an error message that highlights the issue. Plus some tests, because previously this code path was untested.

### DIFF
--- a/loris/img_info.py
+++ b/loris/img_info.py
@@ -78,8 +78,18 @@ class ImageInfo(object):
         self.license = None
         self.service = {}
         self.auth_rules = extra
-        for (k,v) in extra.get('extraInfo', {}).items():
-            setattr(self, k, v)
+
+        # The extraInfo parameter can be used to override specific attributes.
+        # If there are extra attributes, drop an error.
+        bad_attrs = []
+        for (k, v) in extra.get('extraInfo', {}).items():
+            try:
+                setattr(self, k, v)
+            except AttributeError:
+                bad_attrs.append(k)
+        if bad_attrs:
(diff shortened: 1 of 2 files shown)


azaroth42:
Belated :+1:

PR #375, opened 2017-10-17
Add basic flake8 linting in Travis

alexwlchan:
The idea is to catch particularly egregious errors, e.g. undefined variables, before they get into the main branch.

### DIFF
--- a/tests/img_info_t.py
+++ b/tests/img_info_t.py
@@ -461,6 +461,6 @@ def suite():
     test_suites = []
     test_suites.append(unittest.makeSuite(InfoUnit, 'test'))
     test_suites.append(unittest.makeSuite(InfoFunctional, 'test'))
-    test_suites.append(unittest.makeSuite(InfoCache, 'test'))
+    test_suites.append(unittest.makeSuite(TestInfoCache, 'test'))
     test_suite = unittest.TestSuite(test_suites)
     return test_suite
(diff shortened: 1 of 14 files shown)


tests/img_info_t.py
```diff
@@ -461,6 +461,6 @@ def suite():
     test_suites = []
     test_suites.append(unittest.makeSuite(InfoUnit, 'test'))
     test_suites.append(unittest.makeSuite(InfoFunctional, 'test'))
-    test_suites.append(unittest.makeSuite(InfoCache, 'test'))
+    test_suites.append(unittest.makeSuite(TestInfoCache, 'test'))
```
bcail, line 464: Are these lines even necessary anymore? Seems like this fix should result in more tests being run, but it looks like the same number as a previous run, so seems like py.test is just finding the tests without these lines. Do you think we can remove them?

alexwlchan: I don’t use them, I think pytest looks for subclasses of unittest.TestCase and functions/classes whose name includes “Test”. Yeah, I think we can ditch them.

bcail: OK. Let me know if you want to remove any in this PR. Otherwise, I'll merge it and we can remove the lines in a future PR.

alexwlchan: I’ll go ahead and remove them now, it’s only a small extra change.

alexwlchan:
???

I’ve pushed a new commit, but GitHub seems to have lost it.

PR #380, opened 2017-10-20
Add a `safe_rename` function that's atomic and filesystem-aware

### LINKED ISSUE #379: Can not create a cache copy of the image in Centos 7
I am getting below error when the cache is turned on. We are using Centos 7, same configuration in Ubuntu 16.04 seems to be working fine. I tired turning off SELinux but there error still exists. I am using a AWS s3fs mount as the image source root, however changing the image source root  to "/usr/local/share/image" has the same issue. 

Below is the snippet of the error log -
File "/usr/lib/python2.7/site-packages/Loris-2.2.0-py2.7.egg/loris/img.py", line 278, in upsert
rename(temp_fp, target_fp)
OSError: [Errno 18] Invalid cross-device link

alexwlchan:
This uses the proposed approach in #379 – first try to copy the file using `os.rename()`, but if we detect we’re trying to copy across a filesystem boundary and the OS gets upset, we fall back to using `shutil.move()` in a way that’s atomic.

Resolves #379.

/cc @giorgiosironi @bcail

### DIFF
--- a/loris/utils.py
+++ b/loris/utils.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 import errno
 import logging
 import os
+import shutil
 
 
 logger = logging.getLogger(__name__)
@@ -58,4 +59,14 @@ def rename(src, dst):
     So we try ``os.rename()``, but if we detect a cross-filesystem copy, we
     switch to ``shutil.move()`` with some wrappers to make it atomic.
     """
-    os.rename(src, dst)
+    try:
+        os.rename(src, dst)
+    except OSError as err:
+        if err.errno == errno.EXDEV:
+            # First we move it across the filesystem to a temporary file
+            # in a non-atomic way, then move it to the final location
+            # atomically.
+            shutil.move(src, dst + '.tmp')
+            os.rename(dst + '.tmp', dst)
+        else:
+            raise
--- a/loris/utils.py
+++ b/loris/utils.py
@@ -46,17 +46,6 @@ def symlink(src, dst, force=True):
     os.symlink(src, dst)
 
 
-def remove_f(path):
-    """Remove a file at ``path``, but don't error if it doesn't exist."""
-    try:
-        os.remove(path)
-    except OSError as err:
-        if err.errno == errno.ENOENT:
-            pass
-        else:
-            raise
-
-
 def safe_rename(src, dst):
     """Rename a file from ``src`` to ``dst``.
 
--- a/loris/utils.py
+++ b/loris/utils.py
@@ -3,9 +3,11 @@
 from __future__ import absolute_import
 
 import errno
+import glob
 import logging
 import os
 import shutil
+import uuid
 
 
 logger = logging.getLogger(__name__)
@@ -44,6 +46,17 @@ def symlink(src, dst, force=True):
     os.symlink(src, dst)
 
 
+def remove_f(path):
+    """Remove a file at ``path``, but don't error if it doesn't exist."""
+    try:
+        os.remove(path)
+    except OSError as err:
+        if err.errno == errno.ENOENT:
+            pass
+        else:
+            raise
+
+
 def safe_rename(src, dst):
     """Rename a file from ``src`` to ``dst``.
 
@@ -52,6 +65,11 @@ def safe_rename(src, dst):
 
     *   Moves must be atomic.  Otherwise Loris may serve a partial image from
         a cache, which causes an error.  ``shutil.move()`` is not atomic.
+
+        Note that multiple threads may try to write to the cache at once,
+        so atomicity is required to ensure the serving on one thread doesn't
+        pick up a partially saved image from another thread.
+
     *   Moves must work across filesystems.  Often temp directories and the
         cache directories live on different filesystems.  ``os.rename()`` can
         throw errors if run across filesystems.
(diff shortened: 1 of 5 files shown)


alexwlchan:
So it turns out I should probably do something more like this, in case multiple Loris threads try to write to the cache at once: https://stackoverflow.com/a/28090883/1558022

loris/utils.py
```diff
@@ -41,3 +42,31 @@ def symlink(src, dst, force=True):
...
+    switch to ``shutil.move()`` with some wrappers to make it atomic.
+    """
+    try:
+        os.rename(src, dst)
+    except OSError as err:
+        if err.errno == errno.EXDEV:
+            # First we move it across the filesystem to a temporary file
+            # in a non-atomic way, then move it to the final location
+            # atomically.
+            shutil.move(src, dst + '.tmp')
```
giorgiosironi: As you noticed in another comment, using a unique generated name for the temporary file may be safer as different threads/processes could clash on the `.tmp` file. I don't think we need anything more than that because the end product of the different processes is the same content, and the only interaction would be on the `rename` calls, which are atomic.

loris/utils.py
```diff
@@ -41,3 +42,31 @@ def symlink(src, dst, force=True):
...
+    *   Moves must work across filesystems.  Often temp directories and the
+        cache directories live on different filesystems.  ``os.rename()`` can
+        throw errors if run across filesystems.
+
+    So we try ``os.rename()``, but if we detect a cross-filesystem copy, we
+    switch to ``shutil.move()`` with some wrappers to make it atomic.
+    """
+    try:
+        os.rename(src, dst)
+    except OSError as err:
```
giorgiosironi: Perhaps obvious, but there is a performance/complexity penalty of catching the exception in the cross-filesystem case. I don't think it's a problem because the movement of a image file between filesystem will dwarf any local optimization made here.

alexwlchan: Mmm. Assuming we don’t watch to check this every time, we could have an in-memory cache that maps `(src, dst)` to a boolean, recording whether we got an error. If we got an error previously, we skip the `rename(2) `call, and go straight to using `shutil.move()`. Does that sound reasonable?

loris/utils.py
alexwlchan: Yep, and the algorithm I posted (where did that comment go?) had provisions for multiple threads trying to do the copy.

alexwlchan:
Why are computers hard?

Well, it takes 20 lines of code just to copy a file…

loris/utils.py
bcail: do we have any numbers on performance issues with a try/except? Is it worth the complexity of creating a cache? If the try/except is an issue, another possibility would be to add a setting that skips the check and goes right to the shutil.move (might be simpler than maintaining a cache, but of course then the user would have to set the setting to True).

alexwlchan: I don’t like the idea of making it a setting, it’s not something the user should worry about. If we go down that route, we can detect it and cache it – this will only be used on a handful of directories.

I’m busy this week with PyCon UK, but assuming this isn’t too urgent, I can do a bit of perf testing sometime next week.

alexwlchan: I just did some very basic performance testing – rename 10k files from one directory to another, across a filesystem boundary. Two cases:

1. The code as presented here, without a cache
2. Skipping straight into the `shutil.move`, `os.rename` dance – simulating a warm cache that knows not to try `os.rename` in one go.

Time different is negligible, so I think I’ll leave it without a cache.

loris/utils.py
bcail: Looks like this is ready to merge once the tmp file has a unique file name, right?

alexwlchan: It’s a bit more complicated than that, I think, but yes, implementing the whack-a-mole algorithm is the last thing to do.

It's not enough to just give the temp files different names, because threads might interact badly. Consider:

1. Thread A starts copying `src` to `dst-A.tmp`
2. Thread B starts copying `src` to `dst-B.tmp`
3. A finishes, renames `dst-A.tmp` to `dst`, and deletes `src`. A terminates.
4. B is unable to read any more from `src`, so renames the incomplete `dst-B.tmp` to `dst`. B terminates.

And we're left with an incomplete file at `dst`. So B needs to have some smarts to notice if somebody else beat it to the post.

alexwlchan:
rough scribblings before bed (needs comments!):

```python
# https://stackoverflow.com/q/11614815/1558022

import errno
import glob
import os
import shutil
import uuid

def remove_f(path):
    try:
        os.remove(path)
    except OSError as err:
        if err.errno == errno.ENOENT:
            pass
        else:
            raise

if os.path.exists(dst):
    return

mole_id = uuid.uuid4()
tmp_dst = shutil.copyfile(src, '%s.%s.tmp' % (dst, mole_id)))
mole_dst = '%s.%s.mole.tmp' % (dst, mole_id)
os.rename(tmp_dst, mole_dst)

matching = sorted(glob.glob('dst.*.mole.tmp'))

if mole_dst != matching[0]:
    remove_f(mole_dst)

mole_dst = matching[0]

if os.path.exists(dst):
    remove_f(mole_dst)
    return

try:
    os.rename(mole_dst, dst)
except OSError as err:
    if err.errno == errno.ENOENT:
        pass
    else:
        raise

os.unlink(src)
```

loris/utils.py
bcail: I think that both places where you're using safe_rename, the source is a tmp file, which can only have one thread/process writing to it. So I don't think a src getting deleted out from under a process is currently a problem we have to deal with.

If we do want to deal with that problem, how about adding a check before renaming dst-B.tmp to dst? How about something like this?

<pre>
if err.errno == errno.EXDEV:
    f = tempfile.NamedTemporaryFile(dir=..., delete=False)
    shutil.copyfile(src, f.name)
    if os.path.exists(src): #check that source file hasn't already been deleted by another process, so we know we got the whole file
        if not os.path.exists(dst):
            os.rename(f.name, dst)
            os.remove(src)
        else:
            os.remove(f.name) #another process did the copy and will remove the src file
    else:
        os.remove(f.name) #another process already did the copy and removed the src file
else:
    raise
</pre>

loris/utils.py
```diff
@@ -41,3 +44,93 @@ def symlink(src, dst, force=True):
...
+    """
+    logger.debug('Renaming %r to %r', src, dst)
+    try:
+        os.rename(src, dst)
+    except OSError as err:
+        logger.debug('Calling os.rename(%r, %r) failed with %r', src, dst, err)
+        if err.errno == errno.EXDEV:
+
+            # This is based on the safe, atomic file copying algorithm
+            # described in https://stackoverflow.com/a/28090883/1558022.
```
bcail: Isn't some of this unnecessary, because we don't care if the file gets overwritten? If one process creates dst atomically, and then another process overwrites it atomically, is there any harm?

jrhoads:
I may be coming to the conversation late.
Is there any risk of a file starting to be read in between the two atomic
writes?
Or is this vanishingly small?

-Joseph

bcail:
@jrhoads I don't know what the odds are of that happening. If that is an issue (a file being overwritten), I think we'll have the same issue if the directories are on the same filesystem and we can just use the atomic rename (in the first try block of safe_rename). And the issue is already happening in any Loris installs. Does anyone know if overwriting files has been an issue?

bcail:
I ran some tests, and it looks like there's no issue with doing an os.rename of one file onto an identical one while it's being read.

<pre>
import asyncio
import os
import shutil


async def rename_file():
    os.rename('test2.txt', 'test.txt')
    print('  *renamed test2.txt to test.txt*')


async def read_file():
    print('opening file')
    with open('test.txt', 'rt') as f:
        print('reading first 10 bytes')
        first_data = f.read(10)
        await rename_file()
        print('reading rest of file')
        rest_of_data = f.read()
    print('finished reading file - closed it')
    print(f'first_data: {first_data}')
    print(f'rest_of_data: {rest_of_data}')


if __name__ == '__main__':
    assert not os.path.exists('test2.txt')
    shutil.copyfile('test.txt', 'test2.txt')
    print('copied test.txt to test2.txt')
    event_loop = asyncio.get_event_loop()
    try:
        print('starting coroutine')
        coro = read_file()
        print('entering event loop')
        event_loop.run_until_complete(coro)
    finally:
        print('closing event loop')
        event_loop.close()
    assert not os.path.exists('test2.txt')
</pre>

alexwlchan:
> Isn't some of this unnecessary, because we don't care if the file gets overwritten?

Yeah, that version was a bit OTT. I’ve cut it back and pushed a new version.

PR #381, opened 2017-11-02
Allow asterisks in identifiers

giorgiosironi:
One of our image sources is Medium. We found out some of their image filenames contains `*` characters, which is weird when used from a shell but valid.

Originally hot-fixed on our version with https://github.com/elifesciences/loris/commit/06ecff044accc8d6ac5f8d945515acc48450aebe and been in production for several months.

### DIFF
--- a/tests/constants_t.py
+++ b/tests/constants_t.py
@@ -1,20 +1,23 @@
 #-*- coding: utf-8 -*-
 
-import unittest
+import pytest
 
 from loris import constants
 
-class TestImageRequest(unittest.TestCase):
+class TestImageRequest(object):
 
     def test_valid_filenames(self):
         self._assert_valid('/123.jpg/full/full/0/default.jpg')
 
-    def test_valid_special_characters(self):
-        self._assert_valid('/a*b/full/full/0/default.jpg')
-        self._assert_valid('/a:b/full/full/0/default.jpg')
-        self._assert_valid('/a-b/full/full/0/default.jpg')
-        self._assert_valid('/a%b/full/full/0/default.jpg')
+    @pytest.mark.parametrize('path', [
+        '/a*b/full/full/0/default.jpg',
+        '/a:b/full/full/0/default.jpg',
+        '/a-b/full/full/0/default.jpg',
+        '/a%b/full/full/0/default.jpg'
+    ])
+    def test_valid_special_characters(self, path):
+        self._assert_valid(path)
 
     def _assert_valid(self, path):
-        self.assertIsNotNone(constants.IMAGE_RE.match(path))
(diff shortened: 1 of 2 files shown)


alexwlchan:
What behaviour do you see without this patch? Please could you write a test to ensure we don’t break this later?

giorgiosironi:
Essentially file names not allowed by the regex fall into the `redirect_info` request type, and the whole path is treated as an identifier, leading to errors such as:
```
Not Found: could not resolve identifier: lax%3A10627%2Felife-10627-fig1-v1.tif%2Ffull%2Ffull%2F0%2Fdefault.jpg  (404)
```
while only `lax%3A10627%2Felife-10627-fig1-v1.tif` should be treated as the identifier.

giorgiosironi:
The new commit includes tests on several weird filenames. They are directly checked on the constant for simplicity as the immediately above layer would be the whole `loris.webapp` module.

tests/constants_t.py
```diff
@@ -0,0 +1,20 @@
...
+        self._assert_valid('/123.jpg/full/full/0/default.jpg')
+
+    def test_valid_special_characters(self):
+        self._assert_valid('/a*b/full/full/0/default.jpg')
+        self._assert_valid('/a:b/full/full/0/default.jpg')
+        self._assert_valid('/a-b/full/full/0/default.jpg')
+        self._assert_valid('/a%b/full/full/0/default.jpg')
+
+    def _assert_valid(self, path):
+        self.assertIsNotNone(constants.IMAGE_RE.match(path))
```
alexwlchan: Since we’re using pytest for the test runner, you can also use standard assertions. Like so:

```python
assert constants.IMAGE_RE.match(path) is not None
```

tests/constants_t.py
```diff
@@ -0,0 +1,20 @@
...
+import unittest
+
+from loris import constants
+
+class TestImageRequest(unittest.TestCase):
+
+    def test_valid_filenames(self):
+        self._assert_valid('/123.jpg/full/full/0/default.jpg')
+
+    def test_valid_special_characters(self):
```
alexwlchan: Also in the pytest front, it might be nice to use `@pytest.mark.parametrized` here, so that each of these case runs as an isolated test.

alexwlchan (approved):
Thanks, tests make this much nicer. 👌 

A few suggestions about using pytest to make the test code a bit cleaner, but I’d be happy to merge as-is if you don’t want to spend more time on this.

giorgiosironi:
Updated the tests, didn't see the Github notifications for a while.

alexwlchan (approved):
Looks good to me, thanks!

PR #382, opened 2017-11-02
Allow dependancy override

seanwiseman:


alexwlchan:
Thanks for the patch! I’m moderately surprised this isn’t an option in base setuptools, but hey.

Most of the other behaviour in `setup.py` is controlled by command-line flags (see `python setup.py install --help`). I’d be reluctant to add a second way to configure Loris installation – do we use environment variables anywhere else?

seanwiseman:
OK understood, yes I am also surprised.

As you suggested, I think I will add this behaviour via the `cmdclass` you are using in the `setup.py` but for the time being I will just use a workaround on the fork we have.

seanwiseman:
https://github.com/loris-imageserver/loris/pull/382/commits/7612adbaf596b553d8a380ebd17ec62b41705ab7 would be an easy way to simplify the overriding of dependancies (by being able to provide your own requirements.txt with pinned versions). The only thing it removes is the `__import__` checking that was being performed, though I am not sure of the reason for this as if the dependancy is already satisfied on the target system then it would be skipped anyway. Though there may be other reasons for this that I am not aware of.

PR #383, opened 2017-11-17
Ensure that is_resolvable() only returns a boolean

alexwlchan:
We're running Loris with a TemplateHTTPResolver, and we've seen a trickle of 500 errors with the following traceback, caused when a user requests a URL that doesn't match any of our templates:

```
2017-11-17 16:08:21,235 (loris.resolver) [WARNING]: Bad URL request for identifier: 'full/full/0/default.jpg'.
  File "/usr/local/lib/python2.7/dist-packages/Loris-2.0.0-py2.7.egg/loris/webapp.py", line 445, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/local/lib/python2.7/dist-packages/Loris-2.0.0-py2.7.egg/loris/webapp.py", line 388, in wsgi_app
    response = self.route(request)
  File "/usr/local/lib/python2.7/dist-packages/Loris-2.0.0-py2.7.egg/loris/webapp.py", line 408, in route
    if not self.resolver.is_resolvable(ident):
  File "/usr/local/lib/python2.7/dist-packages/Loris-2.0.0-py2.7.egg/loris/resolver.py", line 248, in is_resolvable
    (url, options) = self._web_request_url(ident)
  File "/usr/local/lib/python2.7/dist-packages/Loris-2.0.0-py2.7.egg/loris/resolver.py", line 486, in _web_request_url
    raise ResolverException(404, public_message)
loris.loris_exception.ResolverException: Bad URL request made for identifier: 'full/full/0/default.jpg'.
```

Digging into the error, I found a couple of places where `is_resolvable()` could throw an exception, rather than return a boolean. Because the caller in `webapp.py` expects it to return a boolean, there's no error-handling logic and the exception gets raised as a 500 (despite having a 404 HTTP status).

This patch fixes two sources of those errors, and adds a bunch of tests around this function.

Ready for review.

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -528,21 +528,6 @@ class TemplateHTTPResolver(SimpleHTTPResolver):
             options['verify'] = conf['ssl_check']
         return (url, options)
 
-    def is_resolvable(self, ident):
-        # We can raise a ``ResolverException`` in ``_web_request_url()`` if
-        # somebody tries to look up a URL that doesn't match any of the
-        # templates.  In turn, this causes ``is_resolvable()`` to rethrow
-        # that exception and return a 500 to the user.
-        #
-        # In this case, we should catch the exception and return a boolean.
-        try:
-            return super(TemplateHTTPResolver, self).is_resolvable(ident)
-        except ResolverException as exc:
-            if exc.http_status == 404:
-                return False
-            else:
-                raise
-
 
 class SourceImageCachingResolver(_AbstractResolver):
     '''
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -65,7 +65,7 @@ class _AbstractResolver(object):
             ident (str):
                 The identifier for the image.
         Returns:
-            ImageInfo: Partially constructed ImageInfo object 
+            ImageInfo: Partially constructed ImageInfo object
         Raises:
             ResolverException when something goes wrong...
         """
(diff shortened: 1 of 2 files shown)


loris/resolver.py
```diff
@@ -236,6 +236,7 @@ def request_options(self):
     def is_resolvable(self, ident):
         ident = unquote(ident)
 
+        # TODO: Add some tests around this as well
```
alexwlchan, line 239: This is a future TODO, I don't intend to address it in this patch.

loris/resolver.py
```diff
@@ -517,6 +522,21 @@ def _web_request_url(self, ident):
...
+        #
+        # In this case, we should catch the exception and return a boolean.
+        try:
+            return super(TemplateHTTPResolver, self).is_resolvable(ident)
+        except ResolverException as exc:
+            if exc.http_status == 404:
+                return False
+            else:
+                raise
+
```
bcail: Instead of adding this is_resolvable method, how about catching the ResolverException in SimpleHTTPResolver.is_resolvable? Wouldn't that fix the issue for both resolvers, and not require this new method?

alexwlchan: Ah, I hadn’t noticed that SimpleHTTPResolver can throw a ResolverException in _web_request_url() as well – yeah, I’ll move this up to the parent class.

alexwlchan:
@bcail Okay, I’ve moved the `try … except ResolverException` up into SimpleHTTPResolver, and added the extra tests. Ready for another review.

PR #388, opened 2017-12-02
Move the JP2 Signature/File Type checking into separate methods

alexwlchan:
This is the sort of cleanup alluded to in #387. I’m starting with the first two boxes: the Signature and the File Type. The new methods are fully tested (separating the parser from the disk makes them much easier to test and fuzz), and I’ve added a bunch of explanation and context form the JP2 spec where appropriate.

This implements almost identical logic to the previous check, but failures are more granular, and it’s easier to understand why we're doing these checks.

### DIFF
--- a/loris/jp2_extractor.py
+++ b/loris/jp2_extractor.py
@@ -57,7 +57,7 @@ def _read_jp2_until_match(jp2, match):
         c = struct.unpack('c', b)[0]
         window.append(c)
 
-    jp2.seek(offset=-len(match), whence=os.SEEK_CUR)
+    jp2.seek(-len(match), os.SEEK_CUR)
 
 
 class JP2Extractor(object):
@@ -118,7 +118,7 @@ class JP2Extractor(object):
         # the box was written (see § I.4), so in that case we page forward
         # until we see 'ihdr', which is the start of the next box.
         if file_type_box_length > 0:
-            jp2.read(file_type_box_length - 12)
+            jp2.read(max(file_type_box_length - 12, 0))
         else:
             _read_jp2_until_match(jp2, b'ihdr')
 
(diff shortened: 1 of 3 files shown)


loris/jp2_extractor.py
```diff
@@ -14,31 +23,95 @@ class JP2ExtractionError(LorisException):
...
+
+        file_brand = jp2.read(4)
+        if file_brand != b'jp2\040':
+            raise JP2ExtractionError(
+                "Bad brand in the File Type box: %r" % file_brand
+            )
+
+        # We've already consumed 12 bytes of the box reading the length, type,
+        # and brand fields.  Consume and discard the remaining bytes.
+        jp2.read(file_type_box_length - 12)
```
alexwlchan: This is the one deviation from the previous code – if you look at the next block in `extract_jp2()`, it keeps reading a new byte until it finds the start of the `ihdr` box. Doing it this way allows us to jump straight to the start of the next box.

Note that this may fail if the length is 0 (see § I.4):

> If the value of [the LBox] is 0, then the length of the box was not known when the LBox field was written.

so I may keep the current code that searches for `ihdr`; but it feels unlikely to me that this would be 0 in practice.

bcail: Could you add a check that file_type_box_length - 12 is greater than 0? If not, we could just not do that read at all. Would that work?

alexwlchan: Yeah, I’ve realised this falls over if you call `read()` with a negative number.

So I think I want a compromise: if the length is present, we just use that, otherwise we use the current method to find `ihdr`, then back up 4 bytes. Either way, we end at the beginning of the next box.

bcail: OK.

alexwlchan:
Okay, it correctly handles the case where the box length is 0, and I’ve actually checked in the tests this time! That’s one of the neat things about this sans-IO approach – it’s much easier to test individual cases or fuzz the entire JP2 parser without having a specific file.

Issue #378, opened 2017-10-18
Missing character into constants._IDENT


zannkukai:
Hello,

I'm currently using Loris to serve image from a FedoraCommons repository (3.5)
Checking the documentation, I found the perfect way to use Loris using TemplateHTTPResolver.
For a same object, I can have multiple images. My URL's looks like this
 
> http://myserver.com/fedora/objects/test:12345/datastreams/IMG_0001/content
       http://myserver.com/fedora/objects/test:12345/datastreams/IMG_0002/content
  ...

So I would used this configuration : 

> [resolver]
&nbsp;&nbsp;&nbsp;impl = 'loris.resolver.TemplateHTTPResolver'
&nbsp;&nbsp;&nbsp;cache_root='/usr/local/share/images/loris'
&nbsp;&nbsp;&nbsp;delimiter = "|"
&nbsp;&nbsp;&nbsp;templates = 'fedora_obj_ds'
&nbsp;&nbsp;&nbsp;uri_resolvable=False
&nbsp;&nbsp;&nbsp;head_resolvable=False
[[fedora_obj_ds]]
&nbsp;&nbsp;&nbsp;url = 'http://localhost:8080/fedora/objects/%s/datastreams/%s/content'

So my Loris URL's should be

> /loris/fedora_obj_ds:test:12345|IMG_0001/full/full/0/default.jpg
/loris/fedora_obj_ds:test:12345|IMG_0002/full/full/0/default.jpg

Using the default delimiter "|" mentionned into the loris.conf file, it doesn't work. The LorisRequest.request_type has "redirect_info" value.
Chekcing deeper the code the problem comes from LorisRequest._dissect_uri method. At line 275

`image_match = constants.IMAGE_RE.match(self._path)`

The regexp doesn't match because "|" character isn't allowed into constants.py._IDENT.
Fix this problme is very easy (when we found the problem), adding this character into the constant ::

`(-)  _IDENT = r'(?P<ident>[\w:\-\.\%\/]+)'`
`(+) _IDENT = r'(?P<ident>[\w:|\-\.\%\/]+)'`

Sure, we can use another delimiter, but as "|" is mentionned by default, I think it's easier to update the constant for next release.

Hope it helps

alexwlchan:
Thanks for the report!

> I think it's easier to update the constant for next release.

[...]

PR #389, opened 2017-12-07
allow '|' characters in identifiers (that's the default delimiter for…

bcail:
… TemplateHTTPResolver

### DIFF
--- a/tests/webapp_t.py
+++ b/tests/webapp_t.py
@@ -234,13 +234,13 @@ class TestLorisRequest(TestCase):
         #image request
         path = u'/%s/full/full/0/default.jpg' % identifier
         req = _get_werkzeug_request(path)
-        loris_request = webapp.LorisRequest(req, False, None)
+        loris_request = webapp.LorisRequest(req, redirect_id_slash_to_info=False, proxy_path=None)
         self.assertEqual(loris_request.request_type, u'image')
         self.assertEqual(loris_request.ident, encoded_identifier)
         #info request
         path = u'/%s/info.json' % identifier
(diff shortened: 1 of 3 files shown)


alexwlchan:
Resolves #378.

tests/webapp_t.py
```diff
@@ -228,6 +228,22 @@ def test_many_slash_info_request(self):
...
+        #image request
+        path = u'/%s/full/full/0/default.jpg' % identifier
+        req = _get_werkzeug_request(path)
+        loris_request = webapp.LorisRequest(req, False, None)
+        self.assertEqual(loris_request.request_type, u'image')
+        self.assertEqual(loris_request.ident, encoded_identifier)
+        #info request
+        path = u'/%s/info.json' % identifier
+        req = _get_werkzeug_request(path)
+        loris_request = webapp.LorisRequest(req, False, None)
```
alexwlchan: Minor: please could you use keyword arguments here (and on L237) so it’s clearer what these values mean? i.e.

```python
webapp.LorisRequest(req, redirect_id_slash_to_info=False, proxy_path=None)
```

(Should we consider defaulting those values?)

alexwlchan (approved):
👍 

One comment, but otherwise LGTM. Merge once you’ve added the keyword arguments.

tests/webapp_t.py
bcail: added keyword arguments, and then made default values and removed the arguments from the test completely.

alexwlchan (approved):
👌

PR #394, opened 2018-01-10
Fixing fallback logic in loris.utils.safe_rename

bnbalsamo:
[shutil.copyfile()](https://docs.python.org/2/library/shutil.html#shutil.copyfile) returns None, not a string, resulting in the following call to [os.rename()](https://docs.python.org/2/library/os.html#os.rename) blowing up in the event you end up in this except block.

This fix splits the generation of the tmp_dst string into a separate assignment, and then uses that string in the following calls to both copyfile() and rename()


bcail:
@bnbalsamo Thanks!

Issue #359, opened 2017-09-22
Loris doesn't respect the proxy_path setting when creating the Link header


alexwlchan:
We run Loris with the following setting:

```ini
[loris.Loris]
proxy_path = 'https://iiif.wellcomecollection.org/image/'
```

The proxy_path setting is used to create correct URLs in the info.json, for example:

```console
$ curl 'https://iiif.wellcomecollection.org/image/s3%3AV0000000%2FV0000003.jpg/info.json' 2>/dev/null | jq -r '.["@id"]'
https://iiif.wellcomecollection.org/image/s3%3AV0000000%2FV0000003.jpg
```

But if you request an image file, the Link header is still poking through the origin URL (which in our case, leads to a 404):

```console
$ curl -o /dev/null -D - 'https://iiif.wellcomecollection.org/image/s3%3AV0000000%2FV0000003.jpg/full/full/0/default.jpg' 2>/dev/null | grep Link
Link: <http://iiif.io/api/image/2/level2.json>;rel="profile",<http://iiif-origin.wellcomecollection.org/s3%3AV0000000%2FV0000003.jpg/full/full/0/default.jpg>;rel="canonical"
```

Loris should use the proxy_path setting, and rewrite this URL to the following:

```
http://iiif.wellcomecollection.org/image/s3%3AV0000000%2FV0000003.jpg/full/full/0/default.jpg
```

We’re seeing this while running with commit df047b4.

alexwlchan:
(I’d like to use this bug as a chance to get another person at Wellcome familiar with and contributing to the Loris codebase, so please don’t patch it before next week! 😉 )

bcail:
Fixed in #395.

PR #395, opened 2018-01-29
if defined, use proxy_path for canonical link (issue #359)

bcail:


tests/webapp_t.py
```diff
@@ -380,6 +380,14 @@ def test_image_no_redirect_to_canonical(self):
         resp = self.client.get(to_get, follow_redirects=False)
         self.assertEqual(resp.status_code, 200)
 
+    def test_image_proxy_path_canonical_link(self):
```
alexwlchan, line 383: A second test for the case when we don't have a proxy_path set would be good.

alexwlchan (approved):
Looks good to me, thanks!

One suggestion for better tests, but after that I think this is fine to merge.

PR #398, opened 2018-02-14
[WIP] ExternalAuthorizer

bcail:


bcail:
@azaroth42 I added some tests for the ExternalAuthorizer. The ExternalAuthorizer currently posts information to a "protected_url" in is_protected() - what should the protected url return, to say whether an item is protected or not?

PR #399, opened 2018-02-15
fix img_info tests in py3

bcail:


loris/jp2_extractor.py
```diff
@@ -329,11 +329,11 @@ def extract_jp2(self, jp2):
 
         scaleFactors = []
 
-        window =  deque(jp2.read(2), 2)
+        window = deque(jp2.read(2), 2)
         # start of codestream
-        while map(ord, window) != [0xFF, 0x4F]: # (SOC - required, see pg 14)
+        while ((window[0] != b'\xFF') or (window[1] != b'\x4F')): # (SOC - required, see pg 14)
```
alexwlchan, line 334: Would it be possible to write:

```
while window != b'\xFF\x4F`:
```

I think that's equivalent, isn't it?

bcail: It's not working for me:

In [32]: from collections import deque

In [33]: d = deque(b'\xff\x4f', 2)

In [34]: d == b'\xff\x4f'
Out[34]: False

PR #400, opened 2018-02-15
More py3

bcail:


bcail:
@alexwlchan this PR includes the commit from PR #399.

loris/authorizer.py
```diff
@@ -53,8 +53,8 @@ def __init__(self, config):
 
     def _strip_empty_fields(self, svc):
         # dicts are modified in place
-        for (k, v) in svc.items():
```
alexwlchan, line 56: Does this line not work unmodified on Python 3?

**Edit:** Wait, I see – we're mutating the dict. Never mind.

loris/authorizer.py
```diff
@@ -184,6 +184,8 @@ def __init__(self, config):
 
         self.use_jwt = config.get('use_jwt', True)
         self.salt = config.get('salt', '')
+        if not isinstance(self.salt, bytes):
+            self.salt = self.salt.encode('utf8')
```
alexwlchan: I think we should either drop a warning log that we're encoding the `salt` config option, so there's an explanation if for some reason we've guessed the wrong encoding; or just not encode the string at all, and throw a TypeError instructing the user to change their config.

loris/authorizer.py
```diff
@@ -257,16 +259,21 @@ def _roles_from_request(self, request):
         
         if request.path.endswith("info.json"):
             token = request.headers.get('Authorization', '')        
-            token = token.replace("Bearer", '')
+            if not isinstance(token, bytes):
+                token = token.encode('utf8')
```
alexwlchan: Ditto above.

loris/authorizer.py
```diff
@@ -257,16 +259,21 @@ def _roles_from_request(self, request):
...
+            token = token.replace(b"Bearer", b'')
             cval = token.strip()
             if not cval:
                 return []
-            secret = "%s-%s" % (self.token_secret, origin)
+            secret = b'-'.join([self.token_secret, origin.encode('utf8')])
```
alexwlchan: Rather than encoding it in multiple places, we should do the encode once, at the top of the function.

loris/authorizer.py
```diff
@@ -276,16 +283,17 @@ def _roles_from_request(self, request):
...
             value = fern.decrypt(cval)
-            if not value.startswith(origin):
+            if not value.startswith(origin.encode('utf8')):
                 # Cookie/Token theft
                 return []
             else:
-                value = value[len(origin)+1:]
+                value = value[len(origin.encode('utf8'))+1:]
 
+        if isinstance(value, bytes):
```
alexwlchan: EAFP suggests it might be preferable to write:

```python
try:
    value.decode('utf8')
except AttributeError:
    pass
```

but I don't think it makes a huge amount of difference.

bcail:
@alexwlchan thanks for your review. I've made updates.

loris/authorizer.py
```diff
@@ -251,22 +254,28 @@ def _roles_from_request(self, request):
         origin = request.headers.get('origin', '')
         if not origin:
             origin = request.headers.get('referer', '*')
-        origin = self.basic_origin(origin)
+        origin = self.basic_origin(origin).encode('utf8')
 
-        logger.debug("Got basic origin: %s" % origin)
+        logger.debug('Got basic origin: %s' % origin)
```
alexwlchan: nit, while we're editing this line: logging calls should use commas for string substitution, i.e.

```python
logger.debug('Got basic origin: %s', origin)
```

This avoids doing the string substitution if the log line is never used.

loris/authorizer.py
alexwlchan: Mmm. I think you can iterate over `list(svc.items())` in the same way as before.

Or might it be even simpler just to replace this with a dict comprehension? i.e.

```python
return {k: v for k, v in svc.items() if v}
```

You no longer get in-place modifications, but it feels weird for the function to edit in-place *and* return. It should do one or the other.

loris/authorizer.py
```diff
@@ -203,6 +203,9 @@ def _validate_config(self, config):
                 config['use_jwt']
             )
 
+        if ('salt' in config) and (not isinstance(config['salt'], bytes)):
+            raise ConfigError('"salt" config parameter must be bytes')
```
alexwlchan: Let's include the problem value for easy debugging. Something like:

```python
raise ConfigError(
    '"salt" config parameter must be bytes; got %r (%s)' % (config['salt'], type(config))
)
```

bcail:
@alexwlchan are there any more changes you want me to make before this PR is merged?

alexwlchan:
I’ve been sitting on a half-complete review for days – I’ll try to finish it today or tomorrow.

tests/webapp_t.py
```diff
@@ -278,7 +278,7 @@ class WebappIntegration(loris_t.LorisTest):
     def test_index(self):
         resp = self.client.get('/')
         self.assertEqual(resp.status_code, 200)
-        self.assertTrue(resp.data.startswith('This is Loris, '))
+        self.assertTrue(resp.data.decode('utf8').startswith('This is Loris, '))
```
alexwlchan, line 281: Suggestion: rather than decoding `resp.data`, it might be cleaner to do the comparison with a bytestring:

```python
self.assertTrue(resp.data.startswith(b'This is Loris, '))
```

Ditto all the tests below.

alexwlchan (approved):
Okay, a couple of suggestions, but nothing too major. The improved ConfigError is the only must-address; everything else I'll leave up to you.

Happy for you to merge once you've made the markups. I don't need another review.

PR #407, opened 2018-03-01
Pin the version of Pillow so we can support JPEG-compressed TIFFs, track breaking changes

### LINKED ISSUE #405: Loris seemingly unable to process TIF files
I'm pointing Loris at a Fedora 4.7.2 instance and it's working great using Kakadu for JPEG2000, but I'm seeing the following for TIFFs:

Server Side Error: -2 

(/usr/local/share/images/loris/7b/977/74f/905/03d/4a6/a43/099/a4d/8f5/193/loris_cache.tif). (500)

I'm able to download and render the TIFF from Fedora directly and I doubt it's a permission problem since the permissions look fine and JPEG2000 is working in the image cache. 

I'm guessing it's an issue with Pillow then...

alexwlchan:
This patch adds a test that Loris can handle JPEG-compressed TIFFs correctly.

It also changes the model of storing requirements.txt: rather than a set of loose requirements, we use [pip-tools](https://pypi.org/project/pip-tools/) to turn those into a pinned set of dependencies that we test about. We test with those exact versions, so users can trust they're getting the same dependencies we test with.

When we bump that file, we can retest it in CI and check a dependency hasn't accidentally introduced a breaking change.

Additionally, ensure that running `python setup.py install` gets the same set of requirements you'd get if you `pip install`'d by hand.

Resolves #405.


minusdavid:
Awesome work, @alexwlchan! I woke up to this today and what a great way to wake up!

I think it would be worthwhile to review https://github.com/loris-imageserver/loris/blob/development/doc/dependencies.md as well. 

I realize it's on the development branch... but it's what I used for installing Loris from a release tarball.

I installed Pillow manually following this process. Would that thwart the work you're doing here, or would setup.py warn that Pillow needs to be downgraded if v5.0.0 were already installed on the system?

PR #408, opened 2018-03-02
Remove the "info" attribute from ImageRequest

alexwlchan:
It's bothered me for a while that we have an "info" parameter on ImageRequest which holds an instance of ImageInfo.

A brief reminder of the current process:

* We get a request to serve an image, which is stored in an ImageRequest object
* We resolve the image, then store some metadata in an ImageInfo object
* We put the ImageInfo object on the "info" attribute of ImageRequest

This feels like a weird crossing of boundaries – Request is about information from the user, and Info is about information from the local image store. We're mostly attaching it just as a convenient way to pass this information through the pipeline.

This patch removes this attribute, and passes the Info instance around directly.

As well as being cleaner, it removes code for handling the unusual case where the "info" attribute hasn't been set – our function signatures now enforce that it must exist!

~WIP: Tests aren't passing yet, and I can reduce the churn in the diff when I go back and revisit the naming conventions.~


alexwlchan:
Nearly 50 commits later, I think this is ready for review.

PR #427, opened 2018-06-08
Allow TIFF as an output format

cbutcosk:
Enables output for uncompressed TIFFs. It turns out much of the handling bits are already there--"tif" is in the formats constant so the return content-type is already right as long as the Request objects know it's an allowed format. The base abstract transformer class just needed code to handle making TIFFs, the config needs to know it's an allowed format, and voila.

In principle output could also be compressed (AFAIK PIL understands LZW + JPEG compression as long as libtiff does, it's just a matter of requesting it in the call to `Image.save()` ) but I haven't really taken a close look to see where distros are at with their libtiff support. Anyway responding with a compressed TIFF is not super webby behavior?


bcail:
Thanks!

PR #430, opened 2018-07-07
Improve coverage and tests in resolver.py

alexwlchan:
This started as a small patch, and grew a bit:

* The ident regex code is untested. While adding tests, I realised it can be a standalone piece – so I moved it out into its own class. The new class features a complete set of tests, and caches the regex, so there should be a (minor) performance improvement.

* In the same vein, I moved out the code which constructs cache directory names. I also nixed a pointless unquote-quote cycle, which would at best have cancelled each other out.

* I added a new SimpleHTTPResolver tests which checks if something is resolvable *after* it's been saved to the cache – previously this scenario was untested.

### DIFF
--- a/tests/resolver_t.py
+++ b/tests/resolver_t.py
@@ -294,10 +294,19 @@ class Test_SimpleHTTPResolver(loris_t.LorisTest):
         #Test with a full uri
         #Note: This seems weird but idents resolve wrong and removes a slash from //
         ident = quote_plus('http://sample.sample/0001')
-        expected_path = os.path.join(
-            self.app.resolver.cache_root, 'http',
-            'aa/003/aa5/c26/7b8/633/39a/d76/cd0/30d/c3c', 'loris_cache.tif'
-        )
+        expected_path = join(self.app.resolver.cache_root, 'http')
+        expected_path = join(expected_path, '32')
+        expected_path = join(expected_path, '3a5')
+        expected_path = join(expected_path, '353')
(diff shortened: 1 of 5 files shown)


bcail:
@alexwlchan looks like a couple unused imports failed the flake8 command.

tests/resolver_t.py
```diff
@@ -293,19 +294,10 @@ def test_simple_http_resolver(self):
...
-        expected_path = join(expected_path, '011')
-        expected_path = join(expected_path, '675')
-        expected_path = join(expected_path, 'ebc')
-        expected_path = join(expected_path, 'c75')
-        expected_path = join(expected_path, '083')
-        expected_path = join(expected_path, 'loris_cache.tif')
+        expected_path = os.path.join(
+            self.app.resolver.cache_root, 'http',
+            'aa/003/aa5/c26/7b8/633/39a/d76/cd0/30d/c3c', 'loris_cache.tif'
+        )
```
bcail: why is this expected path different now? Shouldn't the cache paths still resolve to the same place they used to?

alexwlchan: Eesh, looks like I missed an unquote somewhere. Will fix.

bcail:
@alexwlchan looks like there's still an unused import, but feel free to merge once that's fixed.

alexwlchan:
> looks like there's still an unused import, but feel free to merge once that's fixed

Oops, looks like I forgot to push a commit.

PR #437, opened 2018-12-06
Make the use of extra info optional

boosterl:
I wrote a small configuration option (use_extra_info) which makes the use of extra info loris tries to get optional. This is useful when you are using a SimpleHTTPResolver and your server doesn't offer a rules.json file, particularly when it does respond to a request for a rules.json, but returns the image you originally requested and not the accompanying rules.json file, this results in an internal server error.
Because this is really only useful in the SimpleHTTPResolver, I only implemented it there.


alexwlchan:
Thanks for this patch @boosterl! Could you write a test for the change as well, so we don’t inadvertently revert it in a future release?

boosterl:
Hi, so I now added a test for the config option itself, and also a test which checks if a request is still resolvable when the config option is used.

bcail:
Thanks, @boosterl .

PR #442, opened 2019-01-23
Catch hanging openjpeg transforms

bcail:


alexwlchan (commented):
Do you have an example of an OpenJPEG transform that hangs? Could we check that in tests somehow?

bcail:
Well, it could hang if there's a library missing so the executable won't run. Another possibility might be an invalid JP2 image, an image that's not actually JP2. I'll work on some testing.

bcail:
@alexwlchan I rebased this branch, and added a couple tests for terminating an OpenJPEG or Kakadu transform process if the timeout expires.

alexwlchan (approved):
Okay, looks good with tests. Merge when you’re ready. 👍

PR #443, opened 2019-01-23
remove openjpeg binary/lib, and use the packaged libopenjp2-tools

bcail:
@alexwlchan would this work for your development process? The opj_decompress binary I removed was linking to an old png library I don't have.


alexwlchan (approved):
Looks fine to me. I don’t have time to test with my setup right now, but I don’t think it’ll break anything.

If/when I get round to doing more Loris development, I can debug any issues with my setup then.

PR #445, opened 2019-02-25
Allow decimal percentage values for size parameter

lutzhelm:
Loris currently answers with an error (e.g. "Size syntax u'pct:12.5' is not valid.") if the percentage value for the size parameter isn't an integer. But floating point numbers are allowed by the API, and they would work perfectly if only the regex would match them...


bcail:
Thanks!

PR #449, opened 2019-04-25
Setup changes

bcail:
Remove initialization code from setup.py - setup.py just installs the python package. Add a user_commands.py module for users to run setup/init commands after install, if desired.


minusdavid:
Does the change work when installing with pip?

From a packager perspective, this isn't quite good enough. Could you add a BASE_DIR or PREFIX for each of the files created by the "create_default_files_and_directories" function? That way, a packager could specify a package-specific buildroot and then move the files to the vendor specific place they need to go.

minusdavid:
It feels like my suggestion is re-inventing the wheel... so surely there must be other Python web apps that get packaged which have solved this one. 

That said, it looks like other people have grappled with this question:
https://wiki.debian.org/DjangoPackagingDraft

This one is very old but I think was the basis of the Koha Debian package:
https://people.debian.org/~neilm/webapps-policy/

minusdavid:
Seems like Python admits that it's a difficult problem as well:
https://packaging.python.org/overview/

Looking at https://packages.debian.org/stretch/fail2ban, it looks like it uses data_files for /etc and /var files. 

At https://github.com/fail2ban/fail2ban/issues/1536, they also refer to using --root for deployments.

According to some comments at https://ubuntuforums.org/showthread.php?t=1713458, Debian package scripts call the following under the hood:

python setup.py build
python setup.py install --root=debian/package-name/usr --install-layout=deb

The following links speak to that more:
https://www.debian.org/doc/packaging-manuals/python-policy/packaging_tools.html#dh-python
https://manpages.debian.org/testing/python/dh_python2.1.en.html

minusdavid:
The more I look at this... the more I'd have to sit down and spend some time trying out package builds again. Rather than just Googling on a Saturday night... heh.

bcail:
@minusdavid thanks for looking at this. If you want to customize where files are installed, one way to do that with this PR is to customize and pass a python config object into the user_commands.create_default_files_and_directories function. Take a look at the tests/user_commands_t.py module, where I set the function to install the files into a temp directory.

Does that help with creating a package at all? If not, could this PR be a first step, and then we add another option if needed?

minusdavid:
@bcail Thanks for doing the work! 

But no I don't think that would help with creating a package.

But yeah I think this PR would be a first step and then something else could be done to aid with packaging. I don't have the time to invest in this anymore, so I don't want to hold anyone up either.

bcail:
ok, thanks @minusdavid. @alexwlchan any thoughts on this?

loris/user_commands.py
```diff
@@ -0,0 +1,102 @@
...
+
+def _src_code_repo_root():
+    return os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+
+
+def _init_dir(d):
+    # Could do something here to warn if dir exists but permissions or
+    # ownership aren't sufficient.
+    if not os.path.exists(d):
+        os.makedirs(d)
```
alexwlchan: minor: you can reuse `mkdir_p` from `utils.py`, which does this and will handle the case where the directory pops into existence before you call `makedirs`.

alexwlchan (commented):
Getting this code out of `setup.py` seems a good idea to me.

One suggestion for improving code reuse but otherwise nothing big from me; I don’t know a huge amount about packaging.

loris/user_commands.py
bcail: thanks @alexwlchan. Done.

PR #461, opened 2019-06-07
NGINX instructions

lguariento:
This is a document explaining how to deploy Loris server with NGINX and the uWSGI module.


bcail:
thanks @lguariento.

PR #466, opened 2019-06-07
Update nginx.md

lguariento:


bcail:
@lguariento thanks for these updates. I'll merge the PRs, but for the future, you can make these changes all in your fork, and then open 1 PR. (Or you could open one PR and keep pushing more commits to the same branch.)

PR #467, opened 2019-06-07
Update dependencies.md

lguariento:


doc/dependencies.md
```diff
@@ -31,6 +31,20 @@ Then, install all of the dependencies--note that exact versions may vary dependi
     liblcms2-dev liblcms2-utils libtiff5-dev python-dev libwebp-dev apache2 \
     libapache2-mod-wsgi
 
+For CentOS:
+
+    $ sudo yum install zlib-devel freetype libjpeg-turbo-devel lcms2-utils \ 
+    libtiff-devel libwebp-devel openjpeg2
+    
+    Install liblcms-devel:
```
bcail: Is liblcms-devel really required? Are you able to use Loris without it?

lguariento: I've removed it and Loris appears to still do its job, so a first answer would be 'yes'.

PR #468, opened 2019-07-18
Update resolver.py

dvhassel:
fixes a bug to include auth_rules.  If the "Max" parameter is not specified when using rsplit, the method will return the same as the split() method.  When using filenames with multiple dots in them, eg "HP.2007.13.jp2", rsplit('.')[0] won't return the expected string.  In this case, it will return "HP" instead of "HP.2007.13"

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -88,7 +88,7 @@ class _AbstractResolver(object):
         Returns:
             dict: The dict of information to embed in info.json
         """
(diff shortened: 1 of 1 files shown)


bcail:
@dvhassel thanks! Would you be able to add a test showing the bug?

bcail:
I've been in contact with @dvhassel. I'll merge this, and then add a test for it.

PR #478, opened 2019-10-01
ImageInfo - remove "ident"

bcail:
The "ident" is actually the base_uri, and it can be passed in as an argument when a client wants the full IIIF info.json. Otherwise it's not necessary, and confusingly named.

Also remove "protocol" from ImageInfo - it's a constant that can just be referenced when needed.

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -89,9 +89,6 @@ class _AbstractResolver(object):
         else:
             return {}
 
-    def fix_base_uri(self, base_uri):
-        return base_uri
-
     def format_from_ident(self, ident):
         if ident.rfind('.') != -1:
             extension = ident.split('.')[-1]
@@ -139,7 +136,6 @@ class SimpleFSResolver(_AbstractResolver):
 
         source_fp = self.source_file_path(ident)
         format_ = self.format_from_ident(ident)
-        uri = self.fix_base_uri(base_uri)
         extra = self.get_extra_info(ident, source_fp)
         return ImageInfo(app, source_fp, format_, extra)
 
@@ -373,7 +369,6 @@ class SimpleHTTPResolver(_AbstractResolver):
         if not cached_file_path:
             cached_file_path = self.copy_to_cache(ident)
         format_ = self.get_format(cached_file_path, None)
-        uri = self.fix_base_uri(base_uri)
         if self.use_extra_info:
             extra = self.get_extra_info(ident, cached_file_path)
         else:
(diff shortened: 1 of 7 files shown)


loris/resolver.py
```diff
@@ -141,7 +141,7 @@ def resolve(self, app, ident, base_uri):
         format_ = self.format_from_ident(ident)
         uri = self.fix_base_uri(base_uri)
         extra = self.get_extra_info(ident, source_fp)
-        return ImageInfo(app, uri, source_fp, format_, extra)
+        return ImageInfo(app, source_fp, format_, extra)
```
alexwlchan, line 140: Since you’re ditching the `uri` argument here, you can skip defining it on L142 (and I think eventually we get to delete the `fix_base_uri` method).

And in the other two constructors in this file.

tests/authorizer_t.py
```diff
@@ -40,15 +42,14 @@ def test_is_authorized_is_notimplementederror(self):
...
-# This is mostly pointless, as the values returned are static
 class Test_NullAuthorizer(unittest.TestCase):
 
     def setUp(self):
         ident = "test"
         fp = "img/test.png"
         fmt = "png"
         self.authorizer = NullAuthorizer({})
-        self.info = ImageInfo(None, ident, fp, fmt)
+        self.info = ImageInfo(None, fp, fmt)
```
alexwlchan, line 51: Delete the `ident` variable on L48?

tests/authorizer_t.py
```diff
@@ -63,15 +64,14 @@ def test_get_services_info(self):
...
-# This is also mostly pointless, as the values returned are static
 class Test_NooneAuthorizer(unittest.TestCase):
 
     def setUp(self):
         ident = "test"
         fp = "img/test.png"
         fmt = "png"
         self.authorizer = NooneAuthorizer({})
-        self.info = ImageInfo(None, ident, fp, fmt)
+        self.info = ImageInfo(None, fp, fmt)
```
alexwlchan, line 72: Ditto above.

tests/authorizer_t.py
```diff
@@ -85,17 +85,17 @@ def test_get_services_info(self):
...
         ident = "test"
         fp = "img/test.png"
         fmt = "png"
         self.authorizer = SingleDegradingAuthorizer({})
-        self.badInfo = ImageInfo(None, ident, fp, fmt)
-        self.okayInfo = ImageInfo(None, "67352ccc-d1b0-11e1-89ae-279075081939.jp2",\
-            "img/67352ccc-d1b0-11e1-89ae-279075081939.jp2", "jp2")
+        self.badInfo = ImageInfo(None, fp, fmt)
+        self.okayIdent = "67352ccc-d1b0-11e1-89ae-279075081939.jp2"
+        self.okayInfo = ImageInfo(None, "img/%s" % self.okayIdent, "jp2")
```
alexwlchan, line 95: Ditto.

alexwlchan (commented):
Much cleaner!

bcail:
@alexwlchan thanks, I removed those unused lines.

PR #479, opened 2019-10-02
InfoCache

bcail:
Update InfoCache to key off the image identifier, instead of the request url. Also, use the same naming scheme as the SimpleHttpResolver uses for src images (eg. 12/abc/def/...).

Note: this changes the cache layout, so it would invalidate all the objects in the cache and require the cache to be built up again.

### DIFF
--- a/loris/img_info.py
+++ b/loris/img_info.py
@@ -274,13 +274,12 @@ class InfoCache(object):
     iterators, views, default, update, comparators, etc.
 
     Slots:
-        http_root (str): See below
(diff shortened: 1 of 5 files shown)


bcail:
@alexwlchan ping. Any thoughts on this?

alexwlchan (approved):
Sorry, work's been a bit busier than usual lately. Combining the caches and keying on the identifier is smart, looks good. 👍

bcail:
Np, thanks.

PR #480, opened 2019-10-02
check for Transformers creating empty derivatives, and throw an error…

bcail:
… if it happens

### DIFF
--- a/tests/webapp_t.py
+++ b/tests/webapp_t.py
@@ -4,6 +4,7 @@ import os.path
 from os import path, listdir
 from time import sleep
 from unittest import TestCase
+from unittest.mock import patch
 import re
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/webapp_t.py
```diff
@@ -618,6 +619,13 @@ def _assert_tmp_has_no_files(self):
         any_files = any([path.isfile(path.join(tmp, n)) for n in listdir(tmp)])
         self.assertTrue(not any_files, "There are too many files in %s: %s" % (tmp, any_files))
 
+    @patch('loris.transforms.KakaduJP2Transformer.transform')
+    def test_empty_image_file_error(self, mock_transform):
+        to_get = '/%s/full/300,/0/default.jpg' % (self.test_jp2_color_id,)
+        resp = self.client.get(to_get)
+        self.assertEqual(resp.status_code, 500)
+        self.assertTrue('empty derivative file' in resp.data.decode('utf8'))
```
alexwlchan: I'm not sure we should be exposing this error externally; just logging it somewhere.

alexwlchan (commented):
I'm curious – what are the circumstances when this could happen? Is it something you've seen in the wild?

tests/webapp_t.py
bcail: @alexwlchan should that works for me. I could make just a general server side error.

bcail:
@alexwlchan yes, I saw it in our Loris this week. It can happen if there's a problem with the transform that doesn't raise an exception (note: we might need to improve error checking in transform()). Our issue might have been caused by a missing src image file, or something different related to cache moving/pruning.

tests/webapp_t.py
bcail: done - pushed a new commit.

bcail:
@alexwlchan ping. Are you good with this now, or do you want to discuss more?

alexwlchan:
Would be good to improve error checking in the transform, but happy with this as-is.

PR #482, opened 2019-10-16
InfoCache - if the cache has a src_img_fp, but that file has been rem…

bcail:
…oved from the src image cache, return a cache miss so stale file path isn't used

### DIFF
--- a/loris/img_info.py
+++ b/loris/img_info.py
@@ -297,10 +297,10 @@ class InfoCache:
     def _get_ident_dir_path(self, ident):
         ident = unquote(ident)
         return os.path.join(
-                self.root,
-                CacheNamer.cache_directory_name(ident=ident),
-                ident
-            )
+            self.root,
+            CacheNamer.cache_directory_name(ident=ident),
+            ident
+        )
 
     def _get_info_fp(self, ident):
         return os.path.join(self._get_ident_dir_path(ident), 'info.json')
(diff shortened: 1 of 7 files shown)


bcail:
@alexwlchan I've added some commits to this PR. Any thoughts?

alexwlchan (commented):
Idea looks sensible, will try to review early next week. 👍

loris/img_info.py
```diff
@@ -294,7 +295,12 @@ def __init__(self, root, size=500):
         self._lock = Lock()
 
     def _get_ident_dir_path(self, ident):
-        return os.path.join(self.root, CacheNamer.cache_directory_name(ident=ident))
+        ident = unquote(ident)
+        return os.path.join(
+                self.root,
+                CacheNamer.cache_directory_name(ident=ident),
+                ident
+            )
```
alexwlchan: nit: this should be dedented four spaces

loris/img_info.py
```diff
@@ -325,9 +331,15 @@ def get(self, ident):
 
                 lastmod = datetime.utcfromtimestamp(os.path.getmtime(info_fp))
                 info_and_lastmod = (info, lastmod)
-                logger.debug('Info for %s read from file system', ident)
                 # into mem:
                 self.__setitem__(ident, info, _to_fs=False)
+
+        #check that src image exists if it's in the info
```
alexwlchan: This comment wasn't very clear to me when I first read it. Maybe:

> If a source image is referred to in a cached info.json, check it exists on disk.

Would we expect to have an info.json that refers to a non-existent image? Maybe log a warning in that case?

loris/webapp.py
```diff
@@ -549,39 +549,32 @@ def get_info(self, request, ident, base_uri):
...
+        # cache new info if needed
+        if self.enable_caching:
+            self.info_cache[ident] = info
+            # pick up the timestamp... :()
+            info,last_mod = self.info_cache[ident]
+        else:
+            last_mod = None
 
-            return (info,last_mod)
+        return (info,last_mod)
```
alexwlchan: nit: space after commas

alexwlchan (approved):
Couple of minor suggestions, but nothing big. Happy to merge those once that's addressed.

bcail:
@alexwlchan thanks for your comments - I've addressed them in a new commit.

PR #486, opened 2019-10-29
Transforms

bcail:


loris/transforms.py
```diff
@@ -268,6 +263,12 @@ def _scales_to_reduce_arg(self, image_request, image_info):
             arg = str(reduce_arg)
         return arg
 
+    def _create_fifo(self, tmp):
+        fifo_fp = path.join(tmp, 'fifo.bmp')
+        mkfifo_call = '%s %s' % (self.mkfifo, fifo_fp)
+        subprocess.check_call(mkfifo_call.split())
```
alexwlchan: Rather than joining these into a string and then splitting them, could we pass them as the list that `check_call` expects? i.e.

```python
mkfifo_call = [self.mkfifo, fifo_fp]
subprocess.check_call(mkfifo_call)
```

loris/transforms.py
```diff
@@ -303,6 +303,25 @@ def _run_transform(self, target_fp, image_request, image_info, transform_cmd, fi
...
+                'fifo_fp': fifo_fp
+            }
+        )
+        process.start()
+        process.join(self.transform_timeout)
+        if process.is_alive():
+            logger.info('terminating process for %s, %s',
+                image_info.src_img_fp, target_fp)
+            process.terminate()
+            raise TransformException('JP2 transform process timed out')
```
alexwlchan: Maybe worth logging how long it timed out after? i.e.

```suggestion
            raise TransformException('JP2 transform process timed out after %d seconds', self.transform_timeout)
```

alexwlchan (approved):
Couple of suggestions for improvements while we're editing this part of the code, but otherwise looks fine.

bcail:
@alexwlchan thanks, made the updates.

PR #495, opened 2020-01-02
ImageInfo/auth_rules updates

bcail:
1. ImageInfo - rename extra to auth_rules
2. Don't process logo/attribution/... out of extra["extraInfo"], but
allow passing them into ImageInfo as parameters.
3. In the SimpleHttpResolver, only try to fetch rules.json if
use_auth_rules is True.
4. In resolvers, use_auth_rules is now False by default (this will
require anyone using rules.json to update their configuration).

### DIFF
--- a/loris/img_info.py
+++ b/loris/img_info.py
@@ -80,12 +80,11 @@ class ImageInfo(JP2Extractor):
         auth_rules (dict): extra information about authorization [non IIIF]
 
     '''
-    __slots__ = ('scaleFactors', 'width', 'tiles', 'height',
-        'profile', 'sizes', 'service',
-        'attribution', 'logo', 'license', 'auth_rules',
-        'src_format', 'src_img_fp', 'color_profile_bytes')
+    __slots__ = ('width', 'height', 'scaleFactors', 'sizes', 'tiles',
+        'profile', 'service', 'attribution', 'license', 'logo',
+        'src_img_fp', 'src_format', 'color_profile_bytes', 'auth_rules')
 
-    def __init__(self, app=None, src_img_fp="", src_format="", attribution=None, logo=None, license=None, service=None, auth_rules=None):
+    def __init__(self, app=None, service=None, attribution=None, license=None, logo=None, src_img_fp="", src_format="", auth_rules=None):
         self.src_img_fp = src_img_fp
         self.src_format = src_format
         self.attribution = attribution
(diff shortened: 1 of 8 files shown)


loris/img_info.py
```diff
@@ -57,17 +57,22 @@ def default(self, obj):
...
         tiles: [{}]
-        service (dict): services associated with the image
         profile (Profile): Features supported by the server/available for
             this image
+        service [{}]: optional - services associated with the image
+        attribution [{}]: optional - text that must be shown when content obtained from the Image API service is 
+            displayed or used
+        license []: optional - link to an external resource that describes the license or rights statement under which
+            content obtained from the Image API service may be used
+        logo {}: optional - small image that represents an individual or organization associated with the content
```
alexwlchan, line 75: Could we rearrange the order of the slots/descriptions so they match? And also the parameters to `__init__()`?

loris/resolver.py
```diff
@@ -25,13 +25,13 @@
...
 
 
-class _AbstractResolver(object):
+class _AbstractResolver:
 
     def __init__(self, config):
         self.config = config
         if config:
             self.auth_rules_ext = self.config.get('auth_rules_ext', 'rules.json')
-            self.use_extra_info = self.config.get('use_extra_info', True)
```
alexwlchan, line 34: I don’t know how many people have ever set this field in their config – it would be nice to check for the presence of this field, and drop a warning if so. Otherwise stuff will silently break when they update.

Something like:

```python
if "use_extra_info" in self.config and "use_auth_rules" in self.config:
    raise ConfigError("You cannot set both use_extra_info and use_auth_rules. Please remove use_extra_info from your config.")

if "use_extra_info" in self.config:
    warnings.warn("The use_extra_info field has been renamed to use_auth_rules and will be removed in a future version, please update your config", DeprecationWarning)
    self.config["use_auth_rules"] = self.config["use_extra_info"]
```

This might not be the place for it – maybe `config.py`?

loris/resolver.py
```diff
@@ -80,11 +78,12 @@ def get_extra_info(self, ident, source_fp):
...
         """
+        if not self.use_auth_rules:
+            return {}
         xjsfp = source_fp.rsplit('.', 1)[0] + "." + self.auth_rules_ext
         if exists(xjsfp):
-            fh = open(xjsfp)
-            xjs = json.load(fh)
-            fh.close()
+            with open(xjsfp) as fh:
+                xjs = json.load(fh)
```
alexwlchan: 👍 

We could go one step further, and return straight out from inside the `with` block:

```python
with open(xjsfp) as fh:
    return json.load(fh)
```

bcail:
@alexwlchan thanks, I updated the PR.

PR #504, opened 2020-05-15
Fix HTTP resolver caching for identifiers containing url-encodable characters

jamieparkinson:
The HTTP resolver has been saving `unquote(ident)` keys in the cache, but looking up plain `ident`s. That's meant that, for some identifiers, it's found nothing in the cache until just before it saves - after the resolved file has been downloaded again. 

This PR fixes this behaviour by making sure to perform the `unquote` at the time of calculating the cache key. Since `unquote` is idempotent it doesn't break any existing code.

### DIFF
--- a/loris/resolver.py
+++ b/loris/resolver.py
@@ -277,6 +277,7 @@ class SimpleHTTPResolver(_AbstractResolver):
         return (url, self.request_options())
 
     def cache_dir_path(self, ident):
+        ident = unquote(ident)
         return join(
             self.cache_root,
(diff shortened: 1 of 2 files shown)


alexwlchan:
Good spot! Can I persuade you to write a test for it? 😉

jamieparkinson:
You can, although I have already tried and failed - I'll give it another crack

alexwlchan (approved):
Oops. Good catch, thanks!

PR #509, opened 2020-08-22
Normalize content-type before extention computation in SimpleHTTPResolver

### LINKED ISSUE #508: Normalize content-type before extention computation in SimpleHTTPResolver
in `SimpleHTTPResolver`, when we are saving fie to cache, we are computing file-extension from request's `content-type` header. this extension will be then used in computing `source_format` at time of resolving.

But it is not normalizing `content-type` before dictionary lookup. Thus it doesn't support values like `image/jpeg; charset=UTF-8`, (archive.org responds like that.). We have to normalize it to `image/jpeg` before dictionary lookup.

Here is the place in code:

https://github.com/loris-imageserver/loris/blob/3c641eb90bdde33ad5fe8fd1d39a78d424f4a5d7/loris/resolver.py#L301

damooo:
fixes #508


alexwlchan:
Thanks for raising this issue, and the initial patch!

Two suggestions:

* We should use [cgi.parse_header()](https://docs.python.org/3/library/cgi.html?highlight=parse_header#cgi.parse_header) instead of writing our own code to parse the Content-Type header. It’s more likely to handle other edge cases and be robust against other unexpected Content-Type headers.
* Could you add a test for your change please?

damooo:
It now uses `cgi.parse_header`, test case updated to issue complex content-type

damooo:
Hello @alexwlchan, can you merge this patch, so that issue can be fixed?

tests/simple_http_resolver_ut.py
```diff
@@ -63,14 +63,14 @@ def set_responses(self):
...
         )
         responses.add(
                 responses.GET,
                 self.identifier_url,
                 body='II*\x00\x0c\x00\x00\x00\x80\x00  \x0e\x00\x00\x01\x03\x00\x01\x00\x00\x00\x01\x00\x00\x00\x01\x01\x03\x00\x01\x00\x00\x00\x01\x00\x00\x00\x02\x01\x03\x00\x01\x00\x00\x00\x08\x00\x00\x00\x03\x01\x03\x00\x01\x00\x00\x00\x05\x00\x00\x00\x06\x01\x03\x00\x01\x00\x00\x00\x03\x00\x00\x00\x11\x01\x04\x00\x01\x00\x00\x00\x08\x00\x00\x00\x15\x01\x03\x00\x01\x00\x00\x00\x01\x00\x00\x00\x16\x01\x03\x00\x01\x00\x00\x00\x08\x00\x00\x00\x17\x01\x04\x00\x01\x00\x00\x00\x04\x00\x00\x00\x1a\x01\x05\x00\x01\x00\x00\x00\xba\x00\x00\x00\x1b\x01\x05\x00\x01\x00\x00\x00\xc2\x00\x00\x00\x1c\x01\x03\x00\x01\x00\x00\x00\x01\x00\x00\x00(\x01\x03\x00\x01\x00\x00\x00\x02\x00\x00\x00@\x01\x03\x00\x00\x03\x00\x00\xca\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00H\x00\x00\x00\x01\x00\x00\x00H\x00\x00\x00\x01\x00\x00\x00\xff`\xe6q\x19\x08\x00\x00\x80\t\x00\x00\x80\n\x00\x00\x80\x0b\x00\x00\x80\x0c\x00\x00\x80\r',
                 status=200,
-                content_type='image/tiff'
+                content_type='image/tiff; charset=UTF-8'
```
alexwlchan, line 73: This means we're no longer testing the old behaviour – we know the header `image/tiff; charset=UTF-8` works, but not `image/tiff`. I'd rather have a test suite that covers both.

Can you add that, or would you rather I merge it and add the extra test myself?

damooo: I thought that simple case is just one of sub-cases of complex-headers case. Here it tests with any valid `content-type` in it's generic sense, than to check all possible header values.

I will add special `simple` case quickly if it is required.

damooo: at present, it has one `self.identifier_url`, around which test case runs. should i add another resource\_identifier, to do this check? I seems bit lost in figuring out convention test cases following. I fear, i may make it more intrusive. if you don't mind, can you write the test case best suited with conventions?

PR #516, opened 2020-10-16
Ubuntu 20 docs update

schwemmer:
Add instructions for Ubuntu 20 to docs. Slightly restructure instructions for better clarity. Add paragraph about running setup_directories.py script. Add paragraph about where the generated WSGI file can be found. Fix typos.


bcail:
@schwemmer thanks.

PR #519, opened 2020-11-11
loris2 => loris

bcail:


alexwlchan (approved):
Is it worth documenting somewhere that this path has changed? People could upgrade and suddenly their installation stops working.

Otherwise, looks fine. 👍

bcail:
@alexwlchan sure, we can document this in the next release.

PR #527, opened 2021-02-22
Instructions for installing and deploying Loris on Oracle Linux 7.x and Python virtual environment using Apache and mod_wsgi

lguariento:
This is a set of instructions for getting Loris server up and running from scratch on an Oracle Linux 7.x machine with Python virtual environment.


bcail:
thanks

lguariento:
@bcail Shall we also add a link to it from the main 'installing/deploying' page?

bcail:
sure

PR #529, opened 2021-03-05
Remove named pipe (fifo) requirements

bcail:
Using a named pipe for the transforms might be faster. However, the
named pipe requires the user knowing what command to use to create it
(if the default doesn't work). The named pipe also results in the
transform just hanging if there's an error, instead of returning the
error right away.


alexwlchan (approved):
Looks reasonable to me.

PR #531, opened 2021-03-23
Stop quoting all identifiers

bcail:
They need to be quoted in the request URI, so we need to unquote that,
but we shouldn't be quoting all the identifiers.
https://iiif.io/api/image/3.0/#9-uri-encoding-and-decoding


bcail:
@alexwlchan ping

alexwlchan (approved):
Ah sorry, somehow missed the original email about this. LGTM. 👍

PR #556, opened 2023-09-07
Update requirements

stweil:
- fix some security issues caused by too old Python packages
- fix compatibility issues caused by newer versions of Python packages


stweil:
The "ImportError: cannot import name 'BaseResponse' from 'werkzeug.wrappers'" occurs because BaseResponse and BaseRequest have been deprecated starting werkzeug version 2.1.0. Therefore the requirements were updated to require werkzeug < 2.1.0. It would be better to replace BaseResponse and BaseRequest because old versions of werkzeug have high priority security issues.

